PR #50, opened 2019-03-06
Have make fmt run goimports rather than go fmt

negz:
`goimports` is a superset of `gofmt`. It does everything `gofmt` does, and will also automatically import unambiguous import names referenced within a file, and sort imports.

CC @jbw976 regarding your comment on https://github.com/crossplaneio/crossplane/pull/328. I had thought `make fmt` actually formatted our code, but it appears it operates in more of a linter fashion informing us of poorly formatted code. Either way if we expect folks to run `goimports` rather than `gofmt` it makes sense to switch this out.


negz:
https://github.com/golang/go/issues/21476

It's worth noting that `goimports` does not support the simplify flag used by `gofmt`. This is unfortunate, but `make fmt` is effectively duplicating what our linter now does, and we run the `gosimple` linter (among others) to catch such things.

negz:
@bassam It would be nice (but not necessary) to get this merged before https://github.com/crossplaneio/crossplane/pull/332 so that I can include it in the build submodule update I'll need to do there.

PR #51, opened 2019-03-07
Gate Go builds on the linter

negz:
This PR does two things:

Causes `golint-ci` to output checkstyle XML to a file (as opposed to human readable output to the terminal) when the environment variable `RUNNING_IN_CI` is true. I considered automatically attempting to detect a TTY from within `make`, as well as testing for the presence of the `JENKINS_URL` environment variable. I settled on being as explicit as possible - i.e. depending on the CI runner (in our case Jenkins) to set `RUNNING_IN_CI=true`

Configures `make build` to run `make go.lint` (i.e. `golangci-lint`) instead of running `make go.fmt` and `make go.vet`. `golangci-lint` runs many linters, including `gofmt` and `go vet`, and runs fast. This change means builds will not pass if the linter fails, and means we're not invoking the same linters multiple times. Note that in this context we were running `gofmt` as a linter - i.e. running it in diff mode - rather than actually using it to format our files in place.

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -102,9 +102,17 @@ GO_BIN_DIR := $(abspath $(OUTPUT_DIR)/bin)
 GO_OUT_DIR := $(GO_BIN_DIR)/$(PLATFORM)
 GO_TEST_DIR := $(abspath $(OUTPUT_DIR)/tests)
 GO_TEST_OUTPUT := $(GO_TEST_DIR)/$(PLATFORM)
+GO_LINT_DIR := $(abspath $(OUTPUT_DIR)/lint)
+GO_LINT_OUTPUT := $(GO_LINT_DIR)/$(PLATFORM)
 
 ifeq ($(GOOS),windows)
 GO_OUT_EXT := .exe
+endif
+
+# Detect when we're running in CI and output checkstyle XML at lint time rather
+# than human readable output.
+ifeq ($(RUNNING_IN_CI),true)
+GO_LINT_ARGS := --out-format=checkstyle > $(GO_LINT_OUTPUT)/checkstyle.xml
 endif
 
 # NOTE: the install suffixes are matched with the build container to speed up the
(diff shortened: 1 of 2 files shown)


makelib/golang.mk
```diff
@@ -226,12 +235,12 @@ go.generate:
 # Common Targets
 
 build.init: go.init
-build.check: go.fmt
-build.check.platform: go.vet
+build.check: go.lint
```
bassam: I'm a bit confused why you dropped go.fmt and go.vet in this change.

makelib/golang.mk
```diff
@@ -226,12 +235,12 @@ go.generate:
...
 build.code.platform: go.build
 clean: go.clean
 distclean: go.distclean
-lint: go.lint
+lint.init: go.init
+lint.run: go.lint
```
bassam: I think you would need to modify `common.mk` to hook the new `lint.*` targets in.

makelib/golang.mk
negz: @bassam Please take a look at the PR description - I've covered that there. In short, they're redundant.

makelib/golang.mk
negz: That does not seem to be the case. I've tested this and it seems to work as expected when I run  `make build`, `make build.all` or `make lint` using this branch as a submodule of Crossplane.

negz: To be clear - these aren't new targets - `lint.init` and `lint.run` were both already declared in `common.mk`.

negz:
@bassam PTAL. I don't think the comments you left require changes.

makelib/golang.mk
bassam: Aaah I see them now.

makelib/golang.mk
bassam: doh. ok, awesome.

negz:
This PR will run a linter for any Go project using this build submodule. It's stricter than previous linters we've run, so it's likely it will want you to fix a bunch of linter warnings before you can build your code. You can save a config file as `.golangci.yml` at the root of your repo in order to configure how strict the linter is.

This config file will disable the linter completely. Your code will not be checked in any way (including checking whether it is `gofmt`ed):
```yaml
linters:
  disable-all: true
```

This config is functionally equivalent to the build module before this PR. Only `gofmt` and `go vet` will be run as part of `make build`:
```yaml
linters:
  disable-all: true
  enable:
    - gofmt
    - govet
```

PR #54, opened 2019-03-11
Use dep 0.5.1

negz:
https://github.com/golang/dep/blob/master/CHANGELOG.md#v051


negz:
Tested by running `make distclean && make build`. No issues observed.

PR #55, opened 2019-03-19
Build using Go 1.12

negz:
I've built Crossplane by running `GO_SUPPORTED_VERSIONS=1.12 make build` with Go 1.12.1 and confirmed it starts and runs on my minikube.


Defilan:
I definitely support this. Going to 1.12 would be great!! 

Alt Text

PR #56, opened 2019-04-04
Updating kubebuilder to 1.0.8, new go targets

rbwsam:
- New targets:
  - go.test.unit.nocov
  - go.fmt.simplify
  - kubebuilder.bin

Signed-off-by: Sam Leavens <rbwsam@gmail.com>

### DIFF
--- a/cross/Dockerfile
+++ b/cross/Dockerfile
@@ -45,7 +45,7 @@ RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
 # ------------------------------------------------------------------------------------------------
 # Go support
 RUN GO_VERSION=1.12 && \
-    GO_HASH=b3fcf280ff86558e0559e185b601c9eade0fd24c900b4c63cd14d1d38613e499 && \
+    GO_HASH=750a07fef8579ae4839458701f4df690e0b20b8bcce33b437e4df89c451b6f13 && \
     curl -fsSL https://golang.org/dl/go${GO_VERSION}.linux-amd64.tar.gz -o golang.tar.gz && \
     echo "${GO_HASH}  golang.tar.gz" | sha256sum -c - && \
     tar -C /usr/local -xzf golang.tar.gz && \
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -39,6 +39,7 @@ GO_LDFLAGS ?=
 GO_TAGS ?=
 GO_TEST_FLAGS ?=
 GO_TEST_SUITE ?=
+GO_NOCOV ?=
 
 # ====================================================================================
 # Setup go environment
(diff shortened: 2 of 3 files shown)


rbwsam:
@bassam Can you take a look at this PR? The important part is the second commit, it seems that the golang archive we use was modified upstream so I had to update the hash we use for verification to fix the build process.

cross/Dockerfile
```diff
@@ -45,7 +45,7 @@ RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
 # ------------------------------------------------------------------------------------------------
 # Go support
 RUN GO_VERSION=1.12 && \
-    GO_HASH=b3fcf280ff86558e0559e185b601c9eade0fd24c900b4c63cd14d1d38613e499 && \
```
bassam: is this a new version of go? should the `GO_VERSION` be updated too?

makelib/golang.mk
```diff
@@ -166,6 +166,11 @@ go.test.unit: $(GOJUNIT) $(GOCOVER_COBERTURA)
 	@$(GOCOVER_COBERTURA) < $(GO_TEST_OUTPUT)/coverage.txt > $(GO_TEST_OUTPUT)/coverage.xml
 	@$(OK) go test unit-tests
 
+go.test.unit.nocov:
+	@$(INFO) go test unit-tests without coverage
+	@CGO_ENABLED=0 $(GOHOST) test $(ARGS) $(GO_TEST_FLAGS) $(GO_STATIC_FLAGS) $(GO_PACKAGES) || $(FAIL)
```
bassam: could this be accomplished by adding a new environment variable NOCOV instead of a new target?

makelib/golang.mk
```diff
@@ -197,6 +202,11 @@ go.fmt: $(GOFMT)
 	@gofmt_out=$$($(GOFMT) -s -d -e $(GO_SUBDIRS) $(GO_INTEGRATION_TESTS_SUBDIRS) 2>&1) && [ -z "$${gofmt_out}" ] || (echo "$${gofmt_out}" 1>&2; $(FAIL))
 	@$(OK) go fmt
 
+go.fmt.simplify: $(GOFMT)
+	@$(INFO) gofmt simplify
+	@$(GOFMT) -l -s -w $(GO_SUBDIRS) || $(FAIL)
```
bassam: what about `$(GO_INTEGRATION_TESTS_SUBDIRS)`?

cross/Dockerfile
rbwsam: It's the hash for `1.12`, according to git the hash was not updated when `GO_VERSION` was changed from `1.11` to `1.12`. I checked the [golang downloads page](https://golang.org/dl/) and confirmed that that hash is for `1.11`.

bassam: aah so we were still running 1.11 :-(

makelib/golang.mk
rbwsam: Totally, I'll update this PR so it checks for the `NOCOV` env var in the existing `go.test.unit` target.

makelib/golang.mk
rbwsam: Good point, I'll add that.

bassam:
@rbwsam if you can sign the commits, I'll merge.

rbwsam:
@bassam they are all signed now, thanks.

PR #58, opened 2019-04-23
Ensure vendor exists before running Go tests

negz:
I just ran `make distclean && make test` on Crossplane and the tests failed due to missing vendored dependencies.

Disclaimer: I made this edit via the web interface so I haven't actually run it, but expect it to work.


negz:
Oy, this submodule is a pain.

PR #64, opened 2019-06-24
Run verbose Go tests when running in CI

negz:
It turns out we need verbose output in order to parse the output for JUnit.

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -116,10 +116,12 @@ ifeq ($(GOOS),windows)
 GO_OUT_EXT := .exe
 endif
 
-# Detect when we're running in CI and output checkstyle XML at lint time rather
-# than human readable output.
 ifeq ($(RUNNING_IN_CI),true)
+# Output checkstyle XML rather than human readable output.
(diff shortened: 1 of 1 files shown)


makelib/golang.mk
```diff
@@ -116,10 +116,12 @@ ifeq ($(GOOS),windows)
...
 endif
 
-# Detect when we're running in CI and output checkstyle XML at lint time rather
-# than human readable output.
 ifeq ($(RUNNING_IN_CI),true)
+# Output checkstyle XML rather than human readable output.
 GO_LINT_ARGS := --out-format=checkstyle > $(GO_LINT_OUTPUT)/checkstyle.xml
+
+# Output verbose tests that can be parsed into JUnit XML.
+GO_TEST_FLAGS := -v
```
jbw976: should we use `+=` here instead? e.g. `GO_TEST_FLAGS += -v`.  I see places in the build submodule doing that like https://github.com/upbound/build/blob/master/makelib/golang.mk#L74

negz: Probably - fixed.

negz:
It's a little unfortunate that we need to run tests verbosely in the actual log output (making it harder to identify failing tests) in order to get Junit output (which doesn't seem to be super useful in Jenkins).

PR #66, opened 2019-07-30
adding tools for integration tests

soorena776:
cross Dockerfile is updated to have `kubectl`, `kind`, and `helm` backed to to it. These tools are used to do integration testing in Crossplane

### DIFF
--- a/makelib/k8s_tools.mk
+++ b/makelib/k8s_tools.mk
@@ -1,0 +1,50 @@
+# Copyright 2016 The Upbound Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ====================================================================================
+# Options
+
+# the version of kind to use
+KIND_VERSION ?= v0.4.0
+KIND := $(TOOLS_HOST_DIR)/kind-$(KIND_VERSION)
+
+# the version of kubectl to use
+KUBECTL_VERSION ?= v1.15.0
+KUBECTL := $(TOOLS_HOST_DIR)/kubectl-$(KUBECTL_VERSION)
+
+# ====================================================================================
+# Common Targets
+
+k8s_tools.buildvars:
+	@echo KIND=$(KIND)
+	@echo KUBECTL=$(KUBECTL)
+
+build.vars: k8s_tools.buildvars
+
+# ====================================================================================
+# tools
+
+# kind download and install
+$(KIND):
+	@$(INFO) installing kind $(KIND_VERSION)
+	@curl -fsSLo $(KIND) https://github.com/kubernetes-sigs/kind/releases/download/$(KIND_VERSION)/kind-$(GOHOSTOS)-$(GOHOSTARCH) || $(FAIL)
+	@chmod +x $(KIND) 
+	@$(OK) installing kind $(KIND_VERSION)
+
+# kubectl download and install
+$(KUBECTL):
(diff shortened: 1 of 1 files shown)


cross/Dockerfile
```diff
@@ -88,6 +88,22 @@ RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add - && \
...
+# install kind
+RUN curl -Lo /usr/local/bin/kind https://github.com/kubernetes-sigs/kind/releases/download/v0.4.0/kind-linux-amd64 -s &&\
+    chmod +x /usr/local/bin/kind
+
+# install kubectl 
+RUN curl -Lo /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.15.0/bin/linux/amd64/kubectl -s &&\
+    chmod +x /usr/local/bin/kubectl
+
+# install helm
+RUN cd /tmp &&\
```
bassam: Helm is already installed through our make files: https://github.com/upbound/build/blob/master/makelib/helm.mk#L150 

Do we need to install it here too?

soorena776: I noticed that too. However, it seems that this always downloads helm to the src directory and doesn't take advantage of docker cached image. Is this intended?
If we decided to keep helm as is, shouldn't we also install other tools similarly?

soorena776: removed it from the cross image

makelib/common.mk
```diff
@@ -340,6 +340,10 @@ test:
 	@$(MAKE) test.run
 	@$(MAKE) test.done
 
+# integration tests
+test-integration:
```
displague: I would have assumed that something so very Crossplane specific would not live in this common build project.

But after taking another look at https://github.com/crossplaneio/crossplane/blob/master/Makefile, it does appear that the kitchen sink is inherited from this file.

Inclusion here may benefit the Crossplane project as each provider is broken out into its own repository. :+1:

cross/Dockerfile
```diff
@@ -88,6 +88,14 @@ RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add - && \
     DEBIAN_FRONTEND=noninteractive apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
     rm -f /etc/apt/sources.list.d/yarn.list
 
+# ------------------------------------------------------------------------------------------------
+# install kind
+RUN curl -Lo /usr/local/bin/kind https://github.com/kubernetes-sigs/kind/releases/download/v0.4.0/kind-linux-amd64 -s &&\
```
displague: If these versions (0.4.0, 1.15.0) are specified as `ARGS`, then pull requests in crossplane (and provider) projects can bump their build requirements to newer versions as needed without requiring changes in this project (and coordination among subordinate projects).

soorena776: The trade off would be less code readability as the versions and the wiring will be kept away from the Dockerfile. How often do you think these versions will be bumped? 

As a separate concern is imagine we'd use the same version of these tools across multiple projects, and going them in one place would be more manageable.

makelib/common.mk
```diff
@@ -340,6 +340,10 @@ test:
...
+	@${ROOT_DIR}/cluster/local/integration_tests.sh
```
bassam: sorry, I meant that we make `test-integration` a target in the Crossplane project, not here. Specifically I would consider the following:

```
test-integration: $(HELM) $(KIND) $(KUBECTL)
    @$(ROOT_DIR)/cluster/local/integration_tests.sh
```

and make the $(KIND) and $(KUBECTL) new targets possibly in `makelib\kubernetes.mk` similar to https://github.com/upbound/build/blob/master/makelib/kubebuilder.mk#L59

cross/Dockerfile
displague: The default values can be provided: https://docs.docker.com/engine/reference/builder/#default-values so the value would be centralized (this repository) but we could drift away from that default in a provider by provider  / branch by branch case.

makelib/common.mk
suskin: +1 for putting the target which calls the integration test script in the Crossplane repo

cross/Dockerfile
soorena776: Seem's like we are not putting tools in the image anyway. Reverting the change in this Dockerfile, and putting tools targets in Crossplane Makefile.

PR #69, opened 2019-08-12
go: support for goimports

jbw976:
- go.imports.fix target to run goimports and update/fix all issues
  with import lines
- automatically call new go.imports.fix target after running
  go.generate to fix any issues with generated code

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -215,6 +215,11 @@ go.imports: $(GOIMPORTS)
 	@goimports_out=$$($(GOIMPORTS) -d -e -local $(GO_PROJECT) $(GO_SUBDIRS) $(GO_INTEGRATION_TESTS_SUBDIRS) 2>&1) && [ -z "$${goimports_out}" ] || (echo "$${goimports_out}" 1>&2; $(FAIL))
 	@$(OK) goimports
 
+go.imports.fix: $(GOIMPORTS)
+	@$(INFO) goimports fix
+	@$(GOIMPORTS) -l -w -local $(GO_PROJECT) $(GO_SUBDIRS) $(GO_INTEGRATION_TESTS_SUBDIRS) || $(FAIL)
+	@$(OK) goimports fix
+
 go.validate: go.vet go.fmt
 
 go.vendor.lite: $(DEP)
@@ -251,6 +256,7 @@ go.generate:
 	@$(INFO) go generate $(PLATFORM)
 	@CGO_ENABLED=0 $(GOHOST) generate $(GO_COMMON_FLAGS) $(GO_PACKAGES) $(GO_INTEGRATION_TEST_PACKAGES) || $(FAIL)
 	@$(OK) go generate $(PLATFORM)
+	@find $(GO_SUBDIRS) $(GO_INTEGRATION_TESTS_SUBDIRS) -type f -name 'zz_generated*' -exec $(GOIMPORTS) -l -w -local $(GO_PROJECT) {} \;
 
 
 .PHONY: go.init go.build go.install go.test.unit go.test.integration go.test.codecov go.lint go.vet go.fmt go.generate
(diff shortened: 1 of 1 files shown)


makelib/golang.mk
```diff
@@ -251,6 +256,7 @@ go.generate:
 	@$(INFO) go generate $(PLATFORM)
 	@CGO_ENABLED=0 $(GOHOST) generate $(GO_COMMON_FLAGS) $(GO_PACKAGES) $(GO_INTEGRATION_TEST_PACKAGES) || $(FAIL)
 	@$(OK) go generate $(PLATFORM)
+	@$(MAKE) go.imports.fix
```
jbw976: my make-fu is weak, is there a better way to make sure that generated code is properly formatted and `goimports` clean?

hasheddan: This seems sensible to me!

makelib/golang.mk
```diff
@@ -284,6 +290,8 @@ Go Targets:
     generate        Runs go code generation.
     fmt             Checks go source code for formatting issues.
     fmt.simplify    Format, simplify, update source files.
+    imports         Checks go source code import lines for issues.
+    imports.fix     Updates go source files to fix issues with import lines.
```
bassam: did you want to add these targets?

makelib/golang.mk
bassam: `go.imports.fix` will run even for non-generated code which seems a bit odd.

makelib/golang.mk
```diff
@@ -284,6 +290,8 @@ Go Targets:
     generate        Runs go code generation.
```
bassam: do you want to say that this will also run go imports?

makelib/golang.mk
jbw976: ah, i thought it was implicit in this help text that the listed targets were prefixed with `go.`, but that is not the case.  I see that there's a "Special Targets" section above that connects these targets to their `go.*` equivalents.

For consistency, I'd like to add these targets as well.

makelib/golang.mk
jbw976: Yes, that would be good to make that clear from this usage/help text.

makelib/golang.mk
jbw976: I do agree that it's odd to have `goimports` run over the entire codebase instead of just the generated code.  I couldn't determine any way to scope `goimports` to only run over generated code (it's not clear exactly what code was generated), so this was the best functional approach I could find.

Any other suggestions for how to ensure arbitrary generated code is always clean?

jbw976:
@bassam, pushed an update to include all targets that are in the help text. PTAL and let me know any suggestions on better ways to scope the automatic `goimports` run after code is generated.

makelib/golang.mk
bassam: one idea would be to do something like:

```
find $(GO_SUBDIRS) -type f -name 'zz_generated_*' -exec $(GOIMPORTS) ... {} \;
```

jbw976: I could try doing that.  My initial concern is that I wanted this to be generic and be able to catch any generated code, instead of only the `zz_generated*` pattern that controller-tools uses.  That may be fine actually though to scope it to those Kubernetes files.

bassam: might be a good convention to use ```zz_generated``` for all generated files (kubebuilder or not).

bassam: I don't feel strongly about this. its just a bit off to run `make generate` and have a bunch of other files also change because there was a surprising `go.imports` run. its minor.

jbw976: It wasn't sitting right with me to run goimports over the entire codebase, so I got it working to use the suggestion of `find zz_generated*` files. PTAL @bassam, thanks!

PR #70, opened 2019-08-13
`fmt` target is currently calling `go.imports`

bstick12:
Currently the `fmt` target is actually calling the `go.imports`. I also noticed that the `fmt.simplify` target is mentioned in the `make help` but doesn't exist.

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -272,7 +272,8 @@ test.run: go.test.unit
 # ====================================================================================
 # Special Targets
 
-fmt: go.imports
+fmt: go.fmt
+fmt.simplify: go.fmt.simplify
 vendor: go.vendor
 vendor.check: go.vendor.check
 vendor.update: go.vendor.update


bassam:
@bstick12 my understanding is that go.fmt and go.imports are interchangeable and we went with go.imports. See https://goinbigdata.com/goimports-vs-gofmt/

bstick12:
The current `go.fmt` implementation is subtly different to the `go.imports`
https://github.com/upbound/build/blob/e5e43aee51fb07c7630abc4e68ef6a3ba6db7de8/makelib/golang.mk#L203-L206

https://github.com/upbound/build/blob/e5e43aee51fb07c7630abc4e68ef6a3ba6db7de8/makelib/golang.mk#L213-L216

The `go.imports` target is specifying the `--local` which changes the behaviour.

I also  see the is a different PR open to  explicitly add `go.import` targets -  https://github.com/upbound/build/pull/69

negz:
(I'm fairly sure this is no longer relevant.)

PR #63, opened 2019-06-12
modules: update go makefile with modules support

hasheddan:
This commit updates the go init portion of the makefile to support go modules. It is in preparation for migrating Crossplane to go modules: [#534](https://github.com/crossplaneio/crossplane/issues/534)

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

PR #74, opened 2019-09-20
Modify the golang.mk to support go modules

soorena776:
Signed-off-by: soorena776 <javad@upbound.io>

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -132,6 +132,16 @@ endif
 GO_COMMON_FLAGS = $(GO_BUILDFLAGS) -tags '$(GO_TAGS)'
 GO_STATIC_FLAGS = $(GO_COMMON_FLAGS) $(GO_PKG_STATIC_FLAGS) -installsuffix static  -ldflags '$(GO_LDFLAGS)'
 
+export GO111MODULE
+
+# switch for go modules
+ifeq ($(GO111MODULE),on)
+
+# set GOPATH to $(GO_PKG_DIR), so that the go modules are installed there, instead of the default $HOME/go/pkg/mod
+export GOPATH=$(abspath $(GO_PKG_DIR))
+
+endif
+
 # ====================================================================================
 # Go Targets
 
@@ -140,7 +150,7 @@ go.init: go.vendor.lite
 		$(ERR) unsupported go version. Please make install one of the following supported version: '$(GO_SUPPORTED_VERSIONS)' ;\
 		exit 1 ;\
 	fi
-	@if [ "$(realpath ../../../..)" !=  "$(realpath $(GOPATH))" ]; then \
+	@if [ "$(GO111MODULE)" != "on" ] && [ "$(realpath ../../../..)" !=  "$(realpath $(GOPATH))" ]; then \
 		$(WARN) the source directory is not relative to the GOPATH at $(GOPATH) or you are you using symlinks. The build might run into issue. Please move the source directory to be at $(GOPATH)/src/$(GO_PROJECT) ;\
 	fi
 
@@ -217,6 +227,25 @@ go.imports.fix: $(GOIMPORTS)
 
 go.validate: go.vet go.fmt
 
+ifeq ($(GO111MODULE),on)
+
+go.vendor.lite go.vendor.check:
+	@$(INFO) verify dependencies have expected content
+	@$(GOHOST) mod verify || $(FAIL)
+	@$(OK) go modules dependencies verified
+
+go.vendor.update:
+	@$(INFO) update go modules
+	@$(GOHOST) get -u ./... || $(FAIL)
+	@$(OK) update go modules
+
+go.vendor:
+	@$(INFO) go mod vendor
+	@$(GOHOST) mod vendor || $(FAIL)
+	@$(OK) go mod vendor
+
+else
+
 go.vendor.lite: $(DEP)
 #	dep ensure blindly updates the whole vendor tree causing everything to be rebuilt. This workaround
 #	will only call dep ensure if the .lock file changes or if the vendor dir is non-existent.
@@ -241,7 +270,13 @@ go.vendor: $(DEP)
 	@$(DEP) ensure || $(FAIL)
 	@$(OK) dep ensure
 
+endif
+
 go.clean:
+	@# `go modules` creates read-only folders under WORK_DIR
+	@# make all folders within WORK_DIR writable, so they can be deleted
+	@if [ -d $(WORK_DIR) ]; then chmod -R +w $(WORK_DIR); fi 
+
 	@rm -fr $(GO_BIN_DIR) $(GO_TEST_DIR)
 
 go.distclean:
@@ -335,20 +370,20 @@ $(GOFMT):
 $(GOIMPORTS):
 	@$(INFO) installing goimports
 	@mkdir -p $(TOOLS_HOST_DIR)/tmp-imports || $(FAIL)
-	@GOPATH=$(TOOLS_HOST_DIR)/tmp-imports GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get -u golang.org/x/tools/cmd/goimports || rm -fr $(TOOLS_HOST_DIR)/tmp-imports || $(FAIL)
+	@GO111MODULE=off GOPATH=$(TOOLS_HOST_DIR)/tmp-imports GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get -u golang.org/x/tools/cmd/goimports || rm -fr $(TOOLS_HOST_DIR)/tmp-imports || $(FAIL)
 	@rm -fr $(TOOLS_HOST_DIR)/tmp-imports
 	@$(OK) installing goimports
 
 $(GOJUNIT):
 	@$(INFO) installing go-junit-report
 	@mkdir -p $(TOOLS_HOST_DIR)/tmp-junit || $(FAIL)
-	@GOPATH=$(TOOLS_HOST_DIR)/tmp-junit GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get github.com/jstemmer/go-junit-report || rm -fr $(TOOLS_HOST_DIR)/tmp-junit || $(FAIL)
+	@GO111MODULE=off GOPATH=$(TOOLS_HOST_DIR)/tmp-junit GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get github.com/jstemmer/go-junit-report || rm -fr $(TOOLS_HOST_DIR)/tmp-junit || $(FAIL)
 	@rm -fr $(TOOLS_HOST_DIR)/tmp-junit
 	@$(OK) installing go-junit-report
 
 $(GOCOVER_COBERTURA):
 	@$(INFO) installing gocover-cobertura
 	@mkdir -p $(TOOLS_HOST_DIR)/tmp-gocover-cobertura || $(FAIL)
-	@GOPATH=$(TOOLS_HOST_DIR)/tmp-gocover-cobertura GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get github.com/t-yuki/gocover-cobertura || rm -fr $(TOOLS_HOST_DIR)/tmp-covcover-cobertura || $(FAIL)
+	@GO111MODULE=off GOPATH=$(TOOLS_HOST_DIR)/tmp-gocover-cobertura GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get github.com/t-yuki/gocover-cobertura || rm -fr $(TOOLS_HOST_DIR)/tmp-covcover-cobertura || $(FAIL)
 	@rm -fr $(TOOLS_HOST_DIR)/tmp-gocover-cobertura
 	@$(OK) installing gocover-cobertura
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -81,6 +81,8 @@ DEP := $(TOOLS_HOST_DIR)/dep-$(DEP_VERSION)
 GOJUNIT := $(TOOLS_HOST_DIR)/go-junit-report
 GOCOVER_COBERTURA := $(TOOLS_HOST_DIR)/gocover-cobertura
 GOIMPORTS := $(TOOLS_HOST_DIR)/goimports
+CONTROLLER_GEN_VERSION ?= v0.2.1
+CONTROLLERGEN := $(TOOLS_HOST_DIR)/controller-gen
 
 GO := go
 GOHOST := GOOS=$(GOHOSTOS) GOARCH=$(GOHOSTARCH) go
@@ -282,11 +284,11 @@ go.clean:
 go.distclean:
 	@rm -rf $(GO_VENDOR_DIR) $(GO_PKG_DIR)
 
-go.generate: $(GOIMPORTS)
+go.generate: $(GOIMPORTS) $(CONTROLLERGEN)
 	@$(INFO) go generate $(PLATFORM)
-	@CGO_ENABLED=0 $(GOHOST) generate $(GO_COMMON_FLAGS) $(GO_PACKAGES) $(GO_INTEGRATION_TEST_PACKAGES) || $(FAIL)
-	@$(OK) go generate $(PLATFORM)
+	@CGO_ENABLED=0 CONTROLLERGEN=$(CONTROLLERGEN) $(GOHOST) generate $(GO_COMMON_FLAGS) $(GO_PACKAGES) $(GO_INTEGRATION_TEST_PACKAGES) || $(FAIL)
 	@find $(GO_SUBDIRS) $(GO_INTEGRATION_TESTS_SUBDIRS) -type f -name 'zz_generated*' -exec $(GOIMPORTS) -l -w -local $(GO_PROJECT) {} \;
+	@$(OK) go generate $(PLATFORM)
 
 
 .PHONY: go.init go.build go.install go.test.unit go.test.integration go.test.codecov go.lint go.vet go.fmt go.generate
@@ -387,3 +389,16 @@ $(GOCOVER_COBERTURA):
 	@GO111MODULE=off GOPATH=$(TOOLS_HOST_DIR)/tmp-gocover-cobertura GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get github.com/t-yuki/gocover-cobertura || rm -fr $(TOOLS_HOST_DIR)/tmp-covcover-cobertura || $(FAIL)
 	@rm -fr $(TOOLS_HOST_DIR)/tmp-gocover-cobertura
 	@$(OK) installing gocover-cobertura
+
+$(CONTROLLERGEN):
+	@$(INFO) installing controller-gen @$(CONTROLLER_GEN_VERSION)
+	@mkdir -p $(TOOLS_HOST_DIR)/tmp-controllergen || $(FAIL)
+
+	@# `go get` only supports versioned go packages when GO111MODULE=on 
+	@#  since $(TOOLS_HOST_DIR) is under $(GO_PROJECT), make a temp folder which has a go.mod, 
+	@#  so that $(GO_PROJECT)/go.mod doesn't get modified because of running `go get`
+	@cd $(TOOLS_HOST_DIR)/tmp-controllergen; rm -f go.mod; GO111MODULE=on $(GOHOST) mod init tmp-controllergen
+	@cd $(TOOLS_HOST_DIR)/tmp-controllergen; GOPATH=$(abspath $(GO_PKG_DIR)) GO111MODULE=on GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_GEN_VERSION) || $(FAIL)
+	@rm -fr $(TOOLS_HOST_DIR)/tmp-controllergen
+
+	@$(OK) installing controller-gen @$(CONTROLLER_GEN_VERSION)
--- a/makelib/kubebuilder.mk
+++ b/makelib/kubebuilder.mk
@@ -17,6 +17,7 @@
 
 # the version of kubebuilder to use
 KUBEBUILDER_VERSION ?= 1.0.8
+CONTROLLER_GEN_VERSION ?= v0.2.1
 KUBEBUILDER := $(TOOLS_HOST_DIR)/kubebuilder-$(KUBEBUILDER_VERSION)
 
 # these are use by the kube builder test harness
@@ -24,6 +25,39 @@ TEST_ASSET_KUBE_APISERVER := $(KUBEBUILDER)/kube-apiserver
 TEST_ASSET_ETCD := $(KUBEBUILDER)/etcd
 TEST_ASSET_KUBECTL := $(KUBEBUILDER)/kubectl
 export TEST_ASSET_KUBE_APISERVER TEST_ASSET_ETCD TEST_ASSET_KUBECTL
+
+# ====================================================================================
+# Setup environment
+
+-include golang.mk
+
+ifeq ($(CRD_DIR),)
+$(error please set CRD_DIR prior to including kubebuilder.mk)
+endif
+
+ifeq ($(CONTROLLERGEN_PATHS),)
+$(error please set CONTROLLERGEN_PATHS prior to including kubebuilder.mk)
+endif
+
+CONTROLLERGEN := $(TOOLS_HOST_DIR)/controller-gen
+
+# ====================================================================================
+# Kubebuilder Targets
+
+# Generate manifests e.g. CRD, RBAC etc.
+kubebuilder.manifests: $(CONTROLLERGEN)
+	@$(INFO) Generating CRD manifests
+	@# first delete the CRD_DIR, to remove the CRDs of types that no longer exist
+	@rm -rf $(CRD_DIR)
+	@$(CONTROLLERGEN) crd:trivialVersions=true paths=$(CONTROLLERGEN_PATHS) output:dir=$(CRD_DIR)
+	@$(OK) Generating CRD manifests
+
+# Generate controller 
+kubebuilder.generate: $(GOIMPORTS) $(CONTROLLERGEN)
+	@$(INFO) go generate $(PLATFORM)
+	@CGO_ENABLED=0 CONTROLLERGEN=$(CONTROLLERGEN) $(GOHOST) generate $(GO_COMMON_FLAGS) $(GO_PACKAGES) $(GO_INTEGRATION_TEST_PACKAGES) || $(FAIL)
+	@find $(GO_SUBDIRS) $(GO_INTEGRATION_TESTS_SUBDIRS) -type f -name 'zz_generated*' -exec $(GOIMPORTS) -l -w -local $(GO_PROJECT) {} \;
+	@$(OK) go generate $(PLATFORM)
 
 # ====================================================================================
 # Common Targets
@@ -35,8 +69,9 @@ test.init: $(KUBEBUILDER)
 
 define KUBEBULDER_HELPTEXT
 Kubebuilder Targets:
-    bin        run kubebuilder binary, pass args by setting ARGS=""
-    codegen    run code generation
+    bin                     run kubebuilder binary, pass args by setting ARGS=""
+    contgen                 Runs go code generation, to execute controller-gen tool
+    manifests               Generates Kubernetes custom resources manifests (e.g. CRDs RBACs, ...)
 
 endef
 export KUBEBULDER_HELPTEXT
@@ -46,10 +81,13 @@ kubebuilder.help:
 
 help-special: kubebuilder.help
 
+contgen: kubebuilder.generate
+manifests: kubebuilder.manifests
+
 kubebuilder.bin: $(KUBEBUILDER)
 	@$(KUBEBUILDER)/kubebuilder $(ARGS)
 
-.PHONY: codegen kubebuilder.help kubebuilder.bin
+.PHONY: kubebuilder.help kubebuilder.bin kubebuilder.generate kubebuilder.manifests
 
 # ====================================================================================
 # tools
@@ -63,4 +101,16 @@ $(KUBEBUILDER):
 	@rm -fr $(TOOLS_HOST_DIR)/tmp
 	@$(OK) installing kubebuilder $(KUBEBUILDER_VERSION)
 
+$(CONTROLLERGEN):
+	@$(INFO) installing controller-gen @$(CONTROLLER_GEN_VERSION)
+	@mkdir -p $(TOOLS_HOST_DIR)/tmp-controllergen || $(FAIL)
+
+	@# `go get` only supports versioned go packages when GO111MODULE=on 
+	@#  since $(TOOLS_HOST_DIR) is under $(GO_PROJECT), make a temp folder which has a go.mod, 
+	@#  so that $(GO_PROJECT)/go.mod doesn't get modified because of running `go get`
+	@cd $(TOOLS_HOST_DIR)/tmp-controllergen; rm -f go.mod; GO111MODULE=on $(GOHOST) mod init tmp-controllergen
+	@cd $(TOOLS_HOST_DIR)/tmp-controllergen; GOPATH=$(abspath $(GO_PKG_DIR)) GO111MODULE=on GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_GEN_VERSION) || $(FAIL)
+	@rm -fr $(TOOLS_HOST_DIR)/tmp-controllergen
+
+	@$(OK) installing controller-gen @$(CONTROLLER_GEN_VERSION)
 
--- a/makelib/kubebuilder.mk
+++ b/makelib/kubebuilder.mk
@@ -17,7 +17,6 @@
 
 # the version of kubebuilder to use
 KUBEBUILDER_VERSION ?= 1.0.8
-CONTROLLER_GEN_VERSION ?= v0.2.1
 KUBEBUILDER := $(TOOLS_HOST_DIR)/kubebuilder-$(KUBEBUILDER_VERSION)
 
 # these are use by the kube builder test harness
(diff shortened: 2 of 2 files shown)


soorena776:
PTAL @jbw976 @negz

negz:
@soorena776 I like the direction, but I have to admit I'm fairly ignorant around how Go modules work (and also pretty bad at `make`). Can you provide some context around why this PR duplicates the existing library rather than updating or replacing it? I'm also curious what steps if any would be required to migrate our projects from `dep` to modules; my understanding is modules pick the oldest possible dependency while dep picks the newest possible?

negz:
Prior art: https://github.com/upbound/build/pull/63

soorena776:
@negz Sorry about not providing an intro earlier! Please read below:

 `golang_with_modules.mk` is a copy of the existing `golang.mk`, in which 
- `go.vendor.check` is renamed to `go.vendor.verify`, since `dep check` alternative is `go mod verify` 
- `go.vendor` and `go.vendor.verify` have been updated to use `go mod` instead of `dep`
- `go.vendor.lite` has been deleted since its essentially the same as `go.vendor` when using `go mod`
- @(DEP) target is deleted, since it's no longer needed
  
Since other projects are still using `golang.mk`, I created a new library and referenced it in the [consuming project](https://github.com/crossplaneio/stack-aws/pull/19/files#diff-b67911656ef5d18c4ae36cb6741b7965R35)

``

bassam:
I don't think its ideal to have to `golang.mk` files. Any way to combine the two?

soorena776:
@bassam I can definitely combine this with the current one. The only concern there is if other repositories still want to stick with `dep`, and need changes in this repo

soorena776:
@bassam modified the existing `golang.mk` instead

makelib/golang.mk
```diff
@@ -73,14 +78,17 @@ ifneq ($(GO_TEST_SUITE),)
...
 
-GOPATH := $(shell go env GOPATH)
+# no GOPATH is set, since go modules are being used
+GOPATH := 
+
+# enable go modules
+export GO111MODULE=on
 
 # setup tools used during the build
-DEP_VERSION=v0.5.1
```
hasheddan: Does this PR remove `dep` support entirely?

makelib/golang.mk
```diff
@@ -73,14 +78,17 @@ ifneq ($(GO_TEST_SUITE),)
...
-DEP := $(TOOLS_HOST_DIR)/dep-$(DEP_VERSION)
 GOJUNIT := $(TOOLS_HOST_DIR)/go-junit-report
 GOCOVER_COBERTURA := $(TOOLS_HOST_DIR)/gocover-cobertura
 GOIMPORTS := $(TOOLS_HOST_DIR)/goimports
+CONTROLLERGEN := $(TOOLS_HOST_DIR)/controller-gen
```
hasheddan: Not sure the `golang.mk` is the right place to have `controller-gen` and `make manifests`. They seem a little more specific so maybe merit their own Makefile? Maybe we have a `k8scontrollers.mk` or something?

makelib/golang.mk
soorena776: Yes. Is there a reason to keep the support for `dep`? 
I initially created a separate file for `go modules` support. See the first commit.

makelib/golang.mk
soorena776: I was thinking about that too. The issue is that `go.generate` depends on `controller-gen` too. Do you think it would make sense to move `go.generate` out of this makefile as well?

makelib/golang.mk
hasheddan: I don't think it should be a separate file necessarily, just would be nice to support `dep` as well for any projects that want to use it. Otherwise, upgrading the `build` submodule version will break their project.

soorena776:
@hasheddan Thanks for the pointer to #63. I wasn't aware of this effort (my bad in searching existing related work) when I started this. The main difference seems to be that you tried to maintain the support for `dep`, whereas I am not fully convinced if it's necessary. 
In addition, there are a few gotchas when using go modules that I addressed:
- the `~/go/pkg/mod` folder is created by go modules and is read only, which causes issues with the current clean up targets. this was tackled by `go.writeable.mod.folder` target
- supporting manifest and code generation after removing `vendor` folder
- undoing the change in `go.mod` when running `go get` command, which happens when installing tools

Please let me know what you think

makelib/golang.mk
hasheddan: hmm I don't think that having `controller-gen` in `go.generate` is necessary as we didn't previously. I would probably remove `manifests` though.

soorena776: I changed the current go generate tagging in the go code so that it points to the controller-gen binary, instead of running its `main.go` from vendor folder

https://github.com/crossplaneio/stack-aws/pull/20/commits/22c2fc211187c96dfb26653977dff5a89fb96379#diff-3f58b311e6905ae34e04e497070f4881R18

bassam:
I'm not sure we can switch all projects using this project all together. It might be prudent to leave `dep` as the default, and and support for go modules as an option `GO_MODULES=1`. the Crossplane projects can start with that turned on, and we can migrate other projects over time.

soorena776:
@bassam I agree, and that is why I initially created a separate file.
I thought about adding a switch to enable `dep`. The problem is that this will make the already hard-to-read `golang.mk` even more convoluted with conditions and an additional switch for dep. Given that once we migrated from `dep` we will never switch back, do you still think adding a switch makes sense? The alternative is to create a separate `golang.mk` and once all projects are switched just simply delete the older file.

bassam:
I took a quick look at the diff and I'm not sure its that bad to support go modules and dep. I think you can contain the changes in different flavors of targets and using variables for command line switches. 

I worry about the two file approach. I think it will be a a while before everyone switches to go modules, and that in the meantime we have two files to maintain.

makelib/golang.mk
soorena776: Modifying the existing `golang.mk`

makelib/golang.mk
soorena776: added `k8s_controllers.mk`

makelib/golang.mk
```diff
@@ -18,6 +18,11 @@
 # The go project including repo name, for example, github.com/rook/rook
 GO_PROJECT ?= $(PROJECT_REPO)
 
+# the packages to be built statically, for example, $(GO_PROJECT)/cmd/mytool
+ifeq ($(GO_STATIC_PACKAGES),)
+$(error please set GO_STATIC_PACKAGES prior to including golang.mk)
+endif
+
```
bassam: why was this re-added? I think it was taken out to support projects that don't build any packages.

makelib/golang.mk
bassam: can this be removed from there then? Also have you seen `kubebuilder.mk`? seems like it overlaps with `k8s_controller.mk` -- I'd recommend combining.

soorena776:
@bassam @hasheddan @jbw976 
PTAL. It would be great if we finish this pr and check it in, as it's blocking other work.

soorena776:
@bassam 
https://github.com/upbound/build/pull/74#discussion_r328862654

The `CONTROLLERGEN` target is used in `go.generate` target. I'm not sure if moving this target out of this file would make sense. Would it make sense to put the `CONTROLLERGEN` in `k8s_controller.mk` and then `include` that in `golang.mk`?

makelib/golang.mk
```diff
@@ -132,22 +133,37 @@ endif
...
+GO_INIT_REQUIRED_TARGETS = go.vendor.lite
+GO_VENDOR_TARGET = go.vendor
+# switch for go modules
+ifeq ($(GO111MODULE),on)
+# no GOPATH is set, since go modules are being used
+export GOPATH= 
+export GO111MODULE=on
+
+GO_STATIC_FLAGS += -mod=vendor
+GO_INIT_REQUIRED_TARGETS = go.mod.verify
```
jbw976: does this just overwrite the value that was set in `GO_INIT_REQUIRED_TARGETS` earlier on line 136?

makelib/k8s_controllers.mk
```diff
@@ -0,0 +1,53 @@
...
+$(error please set CRD_DIR prior to including k8scontrollers.mk)
+endif
+
+# ====================================================================================
+# k8scontrollers Targets
+
+# Generate manifests e.g. CRD, RBAC etc.
+manifests: $(CONTROLLERGEN)
+	@$(INFO) Generating CRD manifests
+	@$(CONTROLLERGEN) crd:trivialVersions=true paths=./aws/... output:dir=$(CRD_DIR)
```
jbw976: I would not expect to have an `aws` path in this common make library file.

makelib/golang.mk
soorena776: yes, this happens when `GO111MODULE=on`

jbw976: ah yes, i missed that it was within an `ifeq`/`endif` block 😅

makelib/k8s_controllers.mk
soorena776: Could catch! do you think we should just put this target in the local `Makefile`? The reason I brought it here was because the other similar target `codegen` is here as well.

I could also receive parameters from the main Makefile

makelib/golang.mk
```diff
@@ -132,22 +132,37 @@ endif
...
 # Go Targets
 
-go.init: go.vendor.lite
+go.init: $(GO_INIT_REQUIRED_TARGETS)
 	@if ! `$(GO) version | grep -q -E '\bgo($(GO_SUPPORTED_VERSIONS))\b'`; then \
 		$(ERR) unsupported go version. Please make install one of the following supported version: '$(GO_SUPPORTED_VERSIONS)' ;\
 		exit 1 ;\
 	fi
-	@if [ "$(realpath ../../../..)" !=  "$(realpath $(GOPATH))" ]; then \
+	@if [ "$(GO111MODULE)" != "on" ] && [ "$(realpath ../../../..)" !=  "$(realpath $(GOPATH))" ]; then \
```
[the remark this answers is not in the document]
bassam: I think you should be using `GO_PKG_DIR` which is writable.

makelib/golang.mk
```diff
@@ -333,22 +362,21 @@ $(GOFMT):
...
-	@$(OK) installing gocover-cobertura
+	@$(MAKE) go.get GOPKG=github.com/jstemmer/go-junit-report
+
+$(GOCOVER_COBERTURA): $(GOJUNIT)
+	@$(MAKE) go.get GOPKG=github.com/t-yuki/gocover-cobertura
+
+
+go.get:
+	@$(INFO) installing $(GOPKG)
+	@GOBIN=$(TOOLS_HOST_DIR) GO111MODULE=on $(GOHOST) get $(GOPKG) || $(FAIL)
```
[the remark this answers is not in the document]
soorena776: The tool actually gets installed in `$(TOOLS_HOST_DIR)`

makelib/golang.mk
```diff
@@ -241,21 +258,33 @@ go.vendor: $(DEP)
...
+
+go.mod.vendor:
+	@$(INFO) go mod vendor
+	@$(GO) mod vendor || $(FAIL)
+	@$(MAKE) go.writeable.mod.folder
+	@$(OK) go mod vendor
+
+# makes sure the pkg/mod folder is writable (otherwise it won't be deleted)
+go.writeable.mod.folder:
+	@echo
```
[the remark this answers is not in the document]
soorena776: this should be removed

makelib/golang.mk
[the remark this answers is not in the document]
soorena776: With Go modules, you don't get to decide where the modules are installed. By default, they always get installed in `${HOME}/go/pkg/mod`, and this folder is marked as readonly.

makelib/kubebuilder.mk
```diff
@@ -63,4 +100,5 @@ $(KUBEBUILDER):
 	@rm -fr $(TOOLS_HOST_DIR)/tmp
 	@$(OK) installing kubebuilder $(KUBEBUILDER_VERSION)
 
-
+$(CONTROLLERGEN):
+	@$(MAKE) go.get GOPKG=sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_GEN_VERSION)
```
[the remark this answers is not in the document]
soorena776: That is currently the case.

makelib/golang.mk
```diff
@@ -241,21 +258,33 @@ go.vendor: $(DEP)
...
+	@$(GO) mod vendor || $(FAIL)
+	@$(MAKE) go.writeable.mod.folder
+	@$(OK) go mod vendor
+
+# makes sure the pkg/mod folder is writable (otherwise it won't be deleted)
+go.writeable.mod.folder:
+	@echo 
+	@#check if go modules is enabled
+ifeq ($(GO111MODULE),on)
+	@test -d $(shell go env GOPATH)/pkg/mod && find $(shell go env GOPATH)/pkg/mod -type d -exec chmod 777 {} \; || echo
```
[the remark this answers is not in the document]
soorena776: `$(shell go env GOPATH)` gets the default GOPATH (even when the GOPATH env variable is not set). Since we are removing the env variable for GOPATH because of go modules, this command gets the default installation of go modules.

makelib/golang.mk
```diff
@@ -277,22 +306,22 @@ fmt: go.imports
...
 Go Targets:
-    generate        Runs go code generation followed by goimports on generated files.
     fmt             Checks go source code for formatting issues.
     fmt.simplify    Format, simplify, update source files.
     imports         Checks go source code import lines for issues.
     imports.fix     Updates go source files to fix issues with import lines.
     vendor          Updates vendor packages.
     vendor.check    Fail the build if vendor packages have changed.
     vendor.update   Update vendor dependencies.
+    go.mod.verify      Verify go module dependencies have expected content.
```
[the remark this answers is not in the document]
soorena776: This is the reason why I created a separate file for go modules in the first iteration. 

`vendor` folder is not relevant in `go modules`, as it is in `dep`.  Even though there is a `go mod vendor` command to populate the `vendor` folder, but the default source of truth for go packages is  `~/go/pkg/mod`

Because of this, I'm not sure if a `vendor` target for go modules is really meaningful

makelib/golang.mk
```diff
@@ -182,6 +198,7 @@ go.test.integration: $(GOJUNIT)
 	@CGO_ENABLED=0 $(GOHOST) test -i $(GO_STATIC_FLAGS) $(GO_INTEGRATION_TEST_PACKAGES) || $(FAIL)
 	@CGO_ENABLED=0 $(GOHOST) test $(GO_TEST_FLAGS) $(GO_STATIC_FLAGS) $(GO_INTEGRATION_TEST_PACKAGES) $(TEST_FILTER_PARAM) 2>&1 | tee $(GO_TEST_OUTPUT)/integration-tests.log || $(FAIL)
 	@cat $(GO_TEST_OUTPUT)/integration-tests.log | $(GOJUNIT) -set-exit-code > $(GO_TEST_OUTPUT)/integration-tests.xml || $(FAIL)
+	@$(MAKE) go.writeable.mod.folder
```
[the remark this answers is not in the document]
soorena776: because I'm trying to use the same test target for all the cases. `GO11MODULE` is verified inside this function

makelib/golang.mk
```diff
@@ -132,22 +132,37 @@ endif
 GO_COMMON_FLAGS = $(GO_BUILDFLAGS) -tags '$(GO_TAGS)'
 GO_STATIC_FLAGS = $(GO_COMMON_FLAGS) $(GO_PKG_STATIC_FLAGS) -installsuffix static  -ldflags '$(GO_LDFLAGS)'
 
+GO_INIT_REQUIRED_TARGETS = go.vendor.lite
+GO_VENDOR_TARGET = go.vendor
+# switch for go modules
+ifeq ($(GO111MODULE),on)
```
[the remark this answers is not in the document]
soorena776: Currently we are setting the `GOPATH` environment variable for the docker container environment
https://github.com/upbound/build/blob/master/run#L151

With that in place, `GO111MODULE=auto` always will translate to `GO111MODULE=off`.

makelib/golang.mk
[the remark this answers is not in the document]
soorena776: Since we are using `GOPATH` to build the packages in the work directory instead of user home directory, ` $(GOPATH) != ""` won't be the case, even when when `GO111MODULE=on`

makelib/golang.mk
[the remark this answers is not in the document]
soorena776: I made sure that in either case of `dep` and `go modules` nothing is installed or copied in users environment.

makelib/golang.mk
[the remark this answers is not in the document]
soorena776: removed the new targets, in favor of existing vendor targets, keeping the semantics

makelib/golang.mk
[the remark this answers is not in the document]
soorena776: removed `go.writable.mod.folder`, and added logic to the `go.clean` target

makelib/golang.mk
```diff
@@ -132,15 +132,37 @@ endif
...
+GO_VENDOR_TARGET = go.vendor
+GO_VENDOR_CHECK_TARGET = go.vendor.check
+GO_VENDOR_UPDATE_TARGET = go.vendor.update
+
+# switch for go modules
+ifeq ($(GO111MODULE),on)
+
+# set GOPATH to $(WORK_DIR), so that the go modules are installed there, instead of $HOME/go/pkg/mod
+
+GO_MOD_DIR := $(abspath $(WORK_DIR)/mod)
```
bassam: Why not just use GO_PKG_DIR? I don't think we need GO_MOD_DIR.

makelib/golang.mk
```diff
@@ -132,15 +132,37 @@ endif
 GO_COMMON_FLAGS = $(GO_BUILDFLAGS) -tags '$(GO_TAGS)'
 GO_STATIC_FLAGS = $(GO_COMMON_FLAGS) $(GO_PKG_STATIC_FLAGS) -installsuffix static  -ldflags '$(GO_LDFLAGS)'
 
+GO_INIT_REQUIRED_TARGETS = go.vendor.lite
+GO_VENDOR_TARGET = go.vendor
+GO_VENDOR_CHECK_TARGET = go.vendor.check
+GO_VENDOR_UPDATE_TARGET = go.vendor.update
```
bassam: these variables seem unnecessary now that you `go mod` and `dep` are mutually exclusive. I think it would be much simpler to just do:

```
go.vendor:
ifeq ($(GO111MODULE),on)
    <go mod stuff>
else
    <go mod stuff>
endif
```

makelib/golang.mk
```diff
@@ -241,21 +263,36 @@ go.vendor: $(DEP)
...
+	@$(GO) get -u ./... || $(FAIL)
+	@$(OK) update go modules
+
+go.mod.vendor:
+	@$(INFO) go mod vendor
+	@$(GO) mod vendor || $(FAIL)
+	@$(OK) go mod vendor
+
 go.clean:
+	@# make all read-only go modules folder writable, so they can be deleted
```
bassam: I don't follow why this needed if everything goes in the WORK_DIR.

makelib/golang.mk
```diff
@@ -241,21 +263,36 @@ go.vendor: $(DEP)
...
 
 go.distclean:
 	@rm -rf $(GO_VENDOR_DIR) $(GO_PKG_DIR)
 
-go.generate: $(GOIMPORTS)
-	@$(INFO) go generate $(PLATFORM)
-	@CGO_ENABLED=0 $(GOHOST) generate $(GO_COMMON_FLAGS) $(GO_PACKAGES) $(GO_INTEGRATION_TEST_PACKAGES) || $(FAIL)
-	@$(OK) go generate $(PLATFORM)
-	@find $(GO_SUBDIRS) $(GO_INTEGRATION_TESTS_SUBDIRS) -type f -name 'zz_generated*' -exec $(GOIMPORTS) -l -w -local $(GO_PROJECT) {} \;
-
```
bassam: I think there are cases where `go generate` outside of kubebuilder scenarios. does this need to be removed?

makelib/golang.mk
```diff
@@ -333,22 +368,27 @@ $(GOFMT):
...
+endif
+	@GOPATH=$(GO_GET_GOPATH) GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get $(GOPKG) || $(FAIL)
+ifneq ($(GO111MODULE),on)
+	@rm -rf $(GO_GET_GOPATH)
+else
+	@# the below command reverts the go.mod modification because of running `go get`
+	@# otherwise the build becomes dirty and won't be published
+	@$(GOHOST) mod tidy
+endif
+	@$(OK) installed $(GOPKG)
```
bassam: nit: newline.

makelib/golang.mk
```diff
@@ -333,22 +368,27 @@ $(GOFMT):
...
+	@$(MAKE) go.get GOPKG=github.com/jstemmer/go-junit-report
+
+$(GOCOVER_COBERTURA): $(GOJUNIT)
+	@$(MAKE) go.get GOPKG=github.com/t-yuki/gocover-cobertura
+
+go.get:
+	@$(INFO) installing $(GOPKG)
+	$(eval GO_GET_GOPATH := $(GOPATH))
+ifneq ($(GO111MODULE),on)
+	$(eval GO_GET_GOPATH := $(TOOLS_HOST_DIR)/tmp)
```
bassam: I'm not following why this is needed when `GO111MODULE=on`.  Why does the following not work for `dep` and `go mod`:
```
@GOPATH=$(TMP_PATH) GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get $(GOPKG) || rm $(TMP_PATH) || $(FAIL)
```

Also not that you need to use a unique `/tmp` path since make could run multiple tool installs in parallel and they will collide.

makelib/golang.mk
```diff
@@ -333,22 +368,27 @@ $(GOFMT):
...
+	@mkdir -p $(GO_GET_GOPATH) || $(FAIL)
+endif
+	@GOPATH=$(GO_GET_GOPATH) GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get $(GOPKG) || $(FAIL)
+ifneq ($(GO111MODULE),on)
+	@rm -rf $(GO_GET_GOPATH)
+else
+	@# the below command reverts the go.mod modification because of running `go get`
+	@# otherwise the build becomes dirty and won't be published
+	@$(GOHOST) mod tidy
```
bassam: is this needed when `GOPATH` is going to a temp directory?

makelib/kubebuilder.mk
```diff
@@ -63,4 +100,13 @@ $(KUBEBUILDER):
...
 	@$(OK) installing kubebuilder $(KUBEBUILDER_VERSION)
 
-
+$(CONTROLLERGEN):
+	@$(INFO) installing controller-gen @$(CONTROLLER_GEN_VERSION)
+	$(eval PKG_PATH := sigs.k8s.io/controller-tools/cmd/controller-gen)
+ifeq ($(GO111MODULE),on)
+	@$(MAKE) go.get GOPKG=$(PKG_PATH)@$(CONTROLLER_GEN_VERSION)
+else
+	@cd vendor/$(PKG_PATH);GOBIN=$(TOOLS_HOST_DIR) go install
```
bassam: this will install things in the user's home directory, no? Also I don't get why there is a conditional on `GO111MODULE`

soorena776: Currently (w/o `go modules`), the `controller-gen` package is tracked by `dep`, and resides in `$(PROJECT_FOLDER)/vendor`. The `@cd vendor/$(PKG_PATH);GOBIN=$(TOOLS_HOST_DIR) go install` builds and installs the binary in `$(TOOLS_HOST_DIR)`

This tool was being used before by calling the `main.go` file directly, which I thought I should create its binary to be compatible with other tools we use. See https://github.com/crossplaneio/stack-aws/pull/20/files#diff-795fc8c20b8a15e3595e13f3119822e4L18 for the proposed change in stack-aws repository.

makelib/golang.mk
soorena776: As long as  the project directory where `go.mod` resides is a parent to `GOPATH`, running `go get PKG` command  adds that pkg definition to `go.mod`.

Note that since the targets and tools in build repository are supposed to be independent from consuming repositories `go.mod`, it shouldn't be assumed that these packages are already part of `go.mod`

makelib/golang.mk
soorena776: if you have `dir1/some/path/dir2/stuff`, where `dir2` is readonly, `rm -rf dir1` will return `Permission denied`

soorena776: This later causes the `make clean` command fail in Jenkins

makelib/golang.mk
soorena776: If you specify a new GOPATH for `go modules` every time running `go get`, it would download all the packages mentioned in the immediate `go.mod` seen up towards the root folder (which will be the `go.mod` in `PRJ_DIR`). This will be very inefficient for downloading tools one by one

makelib/golang.mk
bassam: none of the paths in WORK_DIR should be read only. Do you know which path is failing?

soorena776: The `GOPATH/pkg/mod` is set to readonly by go, when installing modules
https://github.com/golang/go/issues/27455

soorena776: Also https://github.com/golang/go/issues/27161

I tried using `go clean -modcache` suggested [here](https://github.com/golang/go/issues/27161#issuecomment-415213240), but it wont' work since `make clean` is run from the Jenkins environment, which invokes `go` installed in Jenkins (which is outdated)

soorena776: this is the example of Jenkin's complaining
https://jenkinsci.upbound.io/blue/organizations/jenkins/crossplane%2Fstack-aws/detail/PR-20/66/pipeline

makelib/golang.mk
soorena776: the `tmp` folder is used to download `src` folders like below:
```
tmp
│
└───src
    │
    └───tool1
    │
    └───tool2
```

I put the tmp folder in `WORK_DIR`, and removed the `rm -rf tmp` command and the end of `go get`. Its cleanup is delayed until `make clean` is executed

makelib/golang.mk
soorena776: The main reason is the dependency that `go.vendor.lite`, `go.vendor.lite` and `go.vendor.lite` have on `$(DEP)`

If we use the format above, I'm not sure how easily we can associate the dependency of `$(DEP)` to the `<dep stuff>` inside the loop.

makelib/kubebuilder.mk
```diff
@@ -25,6 +26,38 @@ TEST_ASSET_ETCD := $(KUBEBUILDER)/etcd
...
+
+CONTROLLERGEN := $(TOOLS_HOST_DIR)/controller-gen
+
+# ====================================================================================
+# Kubebuilder Targets
+
+# Generate manifests e.g. CRD, RBAC etc.
+kubebuilder.manifests: $(CONTROLLERGEN)
+	@$(INFO) Generating CRD manifests
+	@$(CONTROLLERGEN) crd:trivialVersions=true paths=$(CONTROLLERGEN_PATHS) output:dir=$(CRD_DIR)
```
muvaf: Could you add a `rm -rf $(CRD_DIR)` before generation call? Because in its current state, generation doesn't delete the CRDs that do not exist anymore but only change the existing ones and add the new ones.

makelib/golang.mk
bassam: good point. how about:

```
ifeq ($(GO111MODULE),on)
go.vendor:
    <go mod stuff>
else
go.vendor: $(DEP)
    <dep stuff>
endif
```

I find the variables with target names a bit harder to read.

makelib/golang.mk
bassam: still curious why you can't just use GO_PKG_DIR?

makelib/golang.mk
bassam: newline is still missing

makelib/golang.mk
bassam: the tmp directory is in `$(TOOLS_HOST_DIR)/tmp)` which is `.cache/linux_amd64/tmp` or something like that. `make clean` does not remove it but `make distclean` will. 

I'm really not sure this `/tmp` optimization is worth the hassle. 1) we cache the finally binary in the .cache directory so it not built every time. 2) the prior logic downloaded the source code and did not cache and had little impact on build time.

makelib/golang.mk
bassam: I'm still not following why if we are using a unique `GOPATH` why we need to tidy anything. maybe you can find me today and discuss.

makelib/kubebuilder.mk
bassam: this logic seems specific to the crossplane project then? I'm not seeing what his makefile is making assumptions about the `vendor/` dir of crossplane.

makelib/golang.mk
bassam: I didn't see a response to this one and it was resolved.

soorena776: I put back the `go.generate` target as you suggested, and then resolved. Do you have more suggestions?

makelib/golang.mk
soorena776: It's the directories which are causing issues, whereas `chmod -R ...` recursively updates both files and folders. If the number of files are large (all the files in downloaded go packages), this could take some time. Do you still rather do that?

makelib/golang.mk
soorena776: kept the existing functionality

makelib/kubebuilder.mk
soorena776: changed to install the package regardless of consumer repositories `vendor/` dir

makelib/golang.mk
soorena776: per our discussion, added temporary `go.mod` file in the tmp folder to avoid this

makelib/kubebuilder.mk
soorena776: added the line

makelib/golang.mk
```diff
@@ -335,20 +368,20 @@ $(GOFMT):
 $(GOIMPORTS):
 	@$(INFO) installing goimports
 	@mkdir -p $(TOOLS_HOST_DIR)/tmp-imports || $(FAIL)
-	@GOPATH=$(TOOLS_HOST_DIR)/tmp-imports GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get -u golang.org/x/tools/cmd/goimports || rm -fr $(TOOLS_HOST_DIR)/tmp-imports || $(FAIL)
+	@GO111MODULE=off GOPATH=$(TOOLS_HOST_DIR)/tmp-imports GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get -u golang.org/x/tools/cmd/goimports || rm -fr $(TOOLS_HOST_DIR)/tmp-imports || $(FAIL)
```
bassam: cool. even better.

makelib/golang.mk
```diff
@@ -217,6 +226,25 @@ go.imports.fix: $(GOIMPORTS)
 
 go.validate: go.vet go.fmt
 
+ifeq ($(GO111MODULE),on)
+
+go.vendor.lite go.vendor.check:
+	@$(INFO) verify dependencies have expected content
+	@$(GO) mod verify || $(FAIL)
```
bassam: should this be $(GOHOST)? Same for below.

makelib/golang.mk
```diff
@@ -281,11 +315,10 @@ vendor: go.vendor
 vendor.check: go.vendor.check
 vendor.update: go.vendor.update
 vet: go.vet
-generate codegen: go.generate
 
 define GO_HELPTEXT
 Go Targets:
-    generate        Runs go code generation followed by goimports on generated files.
+    go.generate     Runs go code generation followed by goimports on generated files.
```
bassam: I think we should leave the original target name to not break projects/flow.

makelib/kubebuilder.mk
```diff
@@ -63,4 +101,16 @@ $(KUBEBUILDER):
...
+	@mkdir -p $(TOOLS_HOST_DIR)/tmp-controllergen || $(FAIL)
+
+	@# `go get` only supports versioned go packages when GO111MODULE=on 
+	@#  since $(TOOLS_HOST_DIR) is under $(GO_PROJECT), make a temp folder which has a go.mod, 
+	@#  so that $(GO_PROJECT)/go.mod doesn't get modified because of running `go get`
+	@cd $(TOOLS_HOST_DIR)/tmp-controllergen; rm -f go.mod; GO111MODULE=on $(GOHOST) mod init tmp-controllergen
+	@cd $(TOOLS_HOST_DIR)/tmp-controllergen; GOPATH=$(abspath $(GO_PKG_DIR)) GO111MODULE=on GOBIN=$(TOOLS_HOST_DIR) $(GOHOST) get sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_GEN_VERSION) || $(FAIL)
+	@rm -fr $(TOOLS_HOST_DIR)/tmp-controllergen
+
+	@$(OK) installing controller-gen @$(CONTROLLER_GEN_VERSION)
```
bassam: can you not use the same `@GO111MODULE=off` approach as in golang.mk?

makelib/golang.mk
```diff
@@ -132,6 +132,15 @@ endif
 GO_COMMON_FLAGS = $(GO_BUILDFLAGS) -tags '$(GO_TAGS)'
 GO_STATIC_FLAGS = $(GO_COMMON_FLAGS) $(GO_PKG_STATIC_FLAGS) -installsuffix static  -ldflags '$(GO_LDFLAGS)'
 
+# switch for go modules
+ifeq ($(GO111MODULE),on)
+
+# set GOPATH to $(GO_PKG_DIR), so that the go modules are installed there, instead of the default $HOME/go/pkg/mod
+export GOPATH=$(abspath $(GO_PKG_DIR))
+export GO111MODULE=on
```
bassam: this seems redundant. Maybe just `export GO111MODULE`

makelib/kubebuilder.mk
soorena776: As I have mentioned in the comments a few line earlier, versioned pkgs are only supported when `GO111MODULE=on`

makelib/golang.mk
soorena776: then how should we resolve the target collision with the new `generate` target that we are adding in kubebuilder.mk?

makelib/kubebuilder.mk
soorena776: No, as the comment above mentions versioned go packages do need `GO111MODULE=on`.

soorena776: I thought about downloading the binary directly, but there is no binary distribution for this tool: https://github.com/kubernetes-sigs/controller-tools/releases/tag/v0.2.1

so we do need to build the go src and install it, by using `go get`

makelib/golang.mk
soorena776: this will make the generate targets  in `golang.mk` and `kubebuilder.mk` to collide which results in executing both and produces an error. I'll rename the target in kubebuilder to avoid the collision.

soorena776:
@bassam tested on jenkins for:

using dep:
https://jenkinsci.upbound.io/blue/organizations/jenkins/crossplane%2Fstack-aws/detail/PR-31/2/pipeline/30

using go modules:
https://jenkinsci.upbound.io/blue/organizations/jenkins/crossplane%2Fstack-aws/detail/PR-20/85/pipeline/

jbw976:
awesome, thank you for working through everything to make this a complete and robust solution @soorena776!

PR #77, opened 2019-11-05
added chart museum via basic auth

arturrez:
added chart museum via basic http auth as per https://app.zenhub.com/workspaces/cloud-platform-and-infrastructure-5c61e874f2a1351031f66999/issues/upbound/infrastructure-squad/50

### DIFF
--- a/makelib/helm.mk
+++ b/makelib/helm.mk
@@ -46,6 +46,12 @@ HELM := $(TOOLS_HOST_DIR)/helm-$(HELM_VERSION)
 # remove the leading `v` for helm chart versions
 HELM_CHART_VERSION := $(VERSION:v%=%)
 
+#Chart Museum variables
+#MUSEUM_URL  ?= "https://helm.example.com/" - url for chart museum
+#If the following variables are set HTTP basic auth will be used. More details https://github.com/helm/chartmuseum/blob/master/README.md#basic-auth
+#MUSEUM_USER ?= "helm"
+#MUSEUM_PASS ?= "changeme"
+
 # ====================================================================================
 # Helm Targets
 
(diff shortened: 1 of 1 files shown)


makelib/helm.mk
```diff
@@ -46,6 +46,11 @@ HELM := $(TOOLS_HOST_DIR)/helm-$(HELM_VERSION)
 # remove the leading `v` for helm chart versions
 HELM_CHART_VERSION := $(VERSION:v%=%)
 
+#Chart Museum
```
jbw976: do you want to leave these commented out variables? if so, i'd encourage some more guidance/info about what they are and how they are expected to be used.

note also that this is a public repository, so anything added here should hopefully be generally applicable and relevant.

makelib/helm.mk
```diff
@@ -113,6 +118,16 @@ helm.promote: $(HELM_HOME)
 	@rm -fr $(HELM_TEMP)
 	@$(OK) promoting helm charts
 
+helm.museum: $(HELM_OUTPUT_DIR)/$(1)-$(HELM_CHART_VERSION).tgz
```
jbw976: what exactly is the `$(1)` in this dependency? this one is a bit unintuitive

makelib/helm.mk
```diff
@@ -113,6 +118,16 @@ helm.promote: $(HELM_HOME)
...
+	ifneq ($(MUSEUM_URL),)
+		@$(INFO) pushing helm charts to chart museum
```
jbw976: want to include the museum URL in this INFO statement?

makelib/helm.mk
arturrez: added more info on how it can be used

makelib/helm.mk
arturrez: done

makelib/helm.mk
arturrez: I think I need to redesign the entire thing. it's not that straightforward that I thought

jbw976: can you expand on that thought @arturrez? I'm not sure I have the full context to understand where you think it needs to go.

arturrez:
no I didn't test it even I tested curl command.
looking into it

PR #79, opened 2019-11-07
Remove kubebuilder library

negz:
I've confirmed no active Crossplane projects will be using this (fairly outdated) library once the below PR is merged.

https://github.com/crossplaneio/stack-azure/pull/62


arturrez:
we have lifter https://github.com/upbound/lifter/blob/master/Makefile currently using it

negz:
@arturrez If lifter is the only remaining consumer of this functionality, how would you feel about merging it into lifter's `Makefile`?

arturrez:
there is number of dependencies for makelib/kubebuilder.mk that I will have to take care about.
I would suggest to leave it here for now at least.

negz:
@arturrez Would you still like me to close this PR without merging? It seems like per #81 this functionality may have been added to lifter, which I believe is the only remaining consumer?

arturrez:
After talking to @rbwsam  I moved functionality to lifter makefile last week. So I don't know any other dependencies on kubebuilder.mk preventing you from merging this PR

PR #81, opened 2019-11-21
add controller_gen back

arturrez:
it was moved to another file  in https://github.com/upbound/build/pull/74  and removed in https://github.com/upbound/build/pull/78

we need it for lifter.
It's temporary as lifter will be replaced.

kubebulder.mk is only used by this app


soorena776:
@arturrez thanks for the effort!

As we discussed offline, since these targets are only used in a specific Makefile, I would suggest to copy this file to the repository. #79 is to remove this file, as it's not being used elsewhere.

arturrez:
thank you

PR #83, opened 2019-12-02
Enable goproxy in build/run

turkenh:
This PR enables Go module proxy by setting in build/run by setting `GOPROXY` env var as `https://proxy.golang.org` (the Go proxy from the Go team, which is run by Google)

Using a proxy will speed up the builds and protect from transient errors like [this](https://github.com/dominikh/go-tools/issues/658) with reproducible builds. It will be enabled by default with Go v1.13. More details and advantages of using a proxy can be found here: https://arslan.io/2019/08/02/why-you-should-use-a-go-module-proxy/

I mainly need this for a case wher build fails without GOPROXY during resolving a dependency even it works when GOPROXY is set.

### DIFF
--- a/run
+++ b/run
@@ -144,6 +144,9 @@ rsync_back() {
     fi
 }
 
+# set GOPROXY
+GOPROXY=https://proxy.golang.org
+
 docker run \
     --rm \
     -h ${BUILD_HOST} \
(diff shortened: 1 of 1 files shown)


muvaf:
As far as I see, this only enables the proxy for docker builds. How about adding the same env var to https://github.com/upbound/build/blob/master/makelib/golang.mk so that all of our builds use the proxy?

turkenh:
Two things here:

1. For local builds, one can already enable go-proxy by exporting the env var (e.g. `export GOPROXY=https://proxy.golang.org`) and make it permanent by adding it to its bash/zsh profile. 
2. With Go 1.13 it will be enabled by default

So, I would prefer to leave this to user for local builds.

turkenh:
Upgrading go 1.13 solves the issue as GOPROXY defaults to `https://proxy.golang.org,direct`.

Closing this in favor of #84

PR #84, opened 2020-01-06
Update golangcilint to 1.22.2 and go to 1.13.5

lukeweber:
I was having odd bugs showing up with trying to use helm v3 as a dependency saying it couldn't be found, like: `go: helm.sh/helm/v3@v3.0.0: unknown revision helm.sh/helm/v3.0.0`

Upgrading to go 1.13.5 seems to have resolved the issue.

Additionally to support go 1.13:
- golangcilint failed and needed an upgrade
- go proxy doesn't work with private repos, so I've added github.com/upound/* to GOPRIVATE.

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -51,7 +51,7 @@ GO_LDFLAGS += -s -w
 endif
 
 # supported go versions
-GO_SUPPORTED_VERSIONS ?= 1.7|1.8|1.9|1.10|1.11|1.12
+GO_SUPPORTED_VERSIONS ?= 1.7|1.8|1.9|1.10|1.11|1.12|1.13
 
 # set GOOS and GOARCH
 GOOS := $(OS)
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -97,7 +97,7 @@ endif
 
 # We use a consistent version of golangci-lint to ensure everyone gets the same
 # linters.
-GOLANGCILINT_VERSION := 1.22.2
+GOLANGCILINT_VERSION ?= 1.22.2
 GOLANGCILINT := $(TOOLS_HOST_DIR)/golangci-lint-v$(GOLANGCILINT_VERSION)
 
 GO_BIN_DIR := $(abspath $(OUTPUT_DIR)/bin)
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -97,7 +97,7 @@ endif
 
 # We use a consistent version of golangci-lint to ensure everyone gets the same
 # linters.
-GOLANGCILINT_VERSION := 1.15.0
+GOLANGCILINT_VERSION := 1.22.2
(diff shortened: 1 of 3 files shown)


run
```diff
@@ -144,6 +144,11 @@ rsync_back() {
     fi
 }
 
+# set GOPROXY
+GOPROXY=https://proxy.golang.org
```
turkenh: With go 1.13 we no longer need to explicitly set this, since it defaults to `https://proxy.golang.org,direct`

From https://golang.org/doc/go1.13#modules:

> The GOPROXY environment variable may now be set to a comma-separated list of proxy URLs or the special token direct, and its default value is now https://proxy.golang.org,direct. When resolving a package path to its containing module, the go command will try all candidate module paths on each proxy in the list in succession. An unreachable proxy or HTTP status code other than 404 or 410 terminates the search without consulting the remaining proxies.

IIUC, with go 1.13 default, we may even not need `GOPRIVATE` since after `proxy.golang.org fails`, it will try `direct`

lukeweber: Ok, I've removed your commit from this PR that included the GOPROXY directive.

makelib/golang.mk
```diff
@@ -97,7 +97,7 @@ endif
 
 # We use a consistent version of golangci-lint to ensure everyone gets the same
 # linters.
-GOLANGCILINT_VERSION := 1.15.0
+GOLANGCILINT_VERSION := 1.22.2
```
jbw976: there were some massive perf and resource usage issues for golangci-lint like 1.17.x or something, are all of those solved and this 1.22.2 looks great during CI build runs?

run
```diff
@@ -144,6 +144,9 @@ rsync_back() {
     fi
 }
 
+# set GOPRIVATE
```
jbw976: want to expand this comment to explain why we're setting this?

run
turkenh: What about `GOPRIVATE` ? may be don't need it as well?

lukeweber: Proxy would have to try, then fail, this seems cleaner. I don't know of any public upbound repos at the moment for code dependencies so this seems more appropriate to go direct and skip the proxy in this case.

makelib/golang.mk
lukeweber: 1.15.0 was really broken with go 1.13. I can only say that my build passed with this combo, but can't speak to if that works in all repos. I assume actually we'll catch some bugs with this better linter on this version, so as people update the build sub-module I expect a small amount of bugfixes. @hasheddan did mention that he's been using 1.18 I think, but I've updated the PR to make this GOLANGCILINT_VERSION configurable in case someone has a preferred version.

run
lukeweber: done

makelib/golang.mk
hasheddan: Yep anything 1.18+ has been working fine for me.

PR #85, opened 2020-01-16
Upgrade node version to 12.x

DaveTorbeck:


DaveTorbeck:
@arturrez @lukeweber We need to upgrade the node version on the frontend to 12.x.  Is this the correct way to do so for the docker image?

PR #86, opened 2020-01-26
Add unzip to cross image

soorena776:
Addresses:
- Adding the unzip command to make it possible to unzip zipped files in cross image
- Fixes: upbound/hosted-crossplane-squad#127

### DIFF
--- a/cross/Dockerfile
+++ b/cross/Dockerfile
@@ -30,6 +30,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
     rsync \
     runit \
     sudo \
+    unzip \
     zip && \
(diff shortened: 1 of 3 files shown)


run
```diff
@@ -146,6 +146,8 @@ rsync_back() {
 
 # set GOPRIVATE to skip go-proxy on upbound by default - https://tip.golang.org/cmd/go/#hdr-Module_configuration_for_non_public_modules
 GOPRIVATE="github.com/upbound/*"
+# set GOPROXY
+GOPROXY=https://proxy.golang.org
```
lukeweber: Go proxy is enabled by default in 1.13 and I don’t think this is default value which would be more like https://proxy.golang.org,direct

soorena776: Reverted this.

PR #87, opened 2020-01-27
Add go to our toolchain

lukeweber:
- Download a given go version
- build with that version

Currently we build with a local version, which is causing issues with some repos being up to date while others don't `support` the newest version. This should make things more consistent in the sense that the ./build/run make and make will use the same underlying configured version of go. As an added bonus you could set your own version.

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -70,7 +70,6 @@ ifneq ($(GO_TEST_SUITE),)
 GO_TEST_FLAGS += -run '$(GO_TEST_SUITE)'
 endif
 
-GOPATH := $(shell $(GO) env GOPATH)
 
 # setup tools used during the build
 DEP_VERSION=v0.5.1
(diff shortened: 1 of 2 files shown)


makelib/golang.mk
```diff
@@ -147,15 +147,11 @@ endif
...
-	@if ! `$(GO) version | grep -q -E '\bgo($(GO_SUPPORTED_VERSIONS))\b'`; then \
-		$(ERR) unsupported go version. Please make install one of the following supported version: '$(GO_SUPPORTED_VERSIONS)' ;\
-		exit 1 ;\
-	fi
 	@if [ "$(GO111MODULE)" != "on" ] && [ "$(realpath ../../../..)" !=  "$(realpath $(GOPATH))" ]; then \
 		$(WARN) the source directory is not relative to the GOPATH at $(GOPATH) or you are you using symlinks. The build might run into issue. Please move the source directory to be at $(GOPATH)/src/$(GO_PROJECT) ;\
 	fi
 
-go.build:
+go.build: $(GO)
```
soorena776: this dependency is already conditioned in build.init

lukeweber:
I would really like to have this work, but I'm not sure of the implications of if it breaks our cross-compile, which isn't really a feature we need IMHO. Also I can't seem to track down why build.check  is called before go is initialized so I need more time to work through cleaning this up. Shelving for today.

PR #88, opened 2020-01-31
Remove goimports

negz:
We invoke goimports in two places:

1. Automatically on generated files.
2. When running make imports

1 is bad practice; we should not edit generated code, even using tools. 2 is redundant; make lint invokes goimports via golangci-lint.

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -80,7 +80,6 @@ DEP_VERSION=v0.5.1
 DEP := $(TOOLS_HOST_DIR)/dep-$(DEP_VERSION)
 GOJUNIT := $(TOOLS_HOST_DIR)/go-junit-report
 GOCOVER_COBERTURA := $(TOOLS_HOST_DIR)/gocover-cobertura
-GOIMPORTS := $(TOOLS_HOST_DIR)/goimports
 
 GO := go
 GOHOST := GOOS=$(GOHOSTOS) GOARCH=$(GOHOSTARCH) go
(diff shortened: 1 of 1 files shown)


negz:
Some context on this PR: I just lost an hour due to a bug in goimports. Running make generate  in stack-azure was deleting corev1 imports from generated deepycopy files, despite them being needed by those files. Turns out make generate invokes goimports on generated files, does not use a pinned version of goimports, and the version of goimports I had cached had a bug in it.

negz:
@jbw976 Was it `golangci-lint` that was balking, or some other system? `golangci-lint` _should_ automatically ignore generated files.

jbw976:
@negz I am not actually sure which specific tool it was (or if it was even linting at all), but a simple `make` was failing without the `goimports` call to clean up the generated files.  Did you already try running make with this change to the build submodule and it completed OK?

negz:
@jbw976 I tested it with stack-azure, but I just now tried stack-gcp and can reproduce the issue you were seeing. I've confirmed that if I add the following to `.golangci-lint.yml` there are no issues:

```yaml
run:
  skip-files:
  - "zz_generated\\..+\\.go$"
```

My suggestion would be to add the above to the linter config in each repo - I don't think there's much value in linting generated code. I'll open PRs for that now.

jbw976:
That sounds very reasonable, thank you for checking on that @negz! so we can go ahead and merge this then, and then pick up the new build submodule hash along with updating the `.golangci-lint.yml` config. Thank you! 💪

negz:
Cross-linking some prior art here:

* https://github.com/golang/lint/issues/239 - `go lint` ignores generated files.
* https://github.com/golangci/golangci-lint/blob/4f3e4105/.golangci.example.yml#L34 - `golangci-lint` appears to intend to ignore generated files (but obviously does not in some cases).

PR #90, opened 2020-02-12
Fix 0.0.0 veresions to avoid invalid SemVer 2.0

### LINKED ISSUE #89: Generating invalid SemVer 2.0 versions
When there are no tags in a repo being built with the tools in this repo it uses this line: https://github.com/upbound/build/blob/510ff11957064de1e4b70c994edaa3a22aad6f11/makelib/common.mk#L190.

This yields versions like `0.0.0-388.4809829` and critically `0.0.0-417.0943445`.
The second example is invalid as SemVer 2.0 does not allow leading zeros in a pre-release section. `git describe` deals with this by prefixing the git sha short hash with a `g`.
```
VERSION := $(shell echo "v0.0.0-$$(git rev-list HEAD --count)-$$(git describe --dirty --always)" | sed 's/-/./2' | sed 's/-/./2')
```
should be:
```
VERSION := $(shell echo "v0.0.0-$$(git rev-list HEAD --count)-g$$(git describe --dirty --always)" | sed 's/-/./2' | sed 's/-/./2')
```

thephred:
Resolves #89

### DIFF
--- a/makelib/common.mk
+++ b/makelib/common.mk
@@ -187,7 +187,7 @@ ifeq ($(origin VERSION), undefined)
 # check if there are any existing `git tag` values
 ifeq ($(shell git tag),)
 # no tags found - default to initial tag `v0.0.0`
-VERSION := $(shell echo "v0.0.0-$$(git rev-list HEAD --count)-$$(git describe --dirty --always)" | sed 's/-/./2' | sed 's/-/./2')
+VERSION := $(shell echo "v0.0.0-$$(git rev-list HEAD --count)-g$$(git describe --dirty --always)" | sed 's/-/./2' | sed 's/-/./2')
 else
 # use tags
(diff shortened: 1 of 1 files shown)


makelib/common.mk
```diff
@@ -187,7 +187,7 @@ ifeq ($(origin VERSION), undefined)
 # check if there are any existing `git tag` values
 ifeq ($(shell git tag),)
 # no tags found - default to initial tag `v0.0.0`
-VERSION := $(shell echo "v0.0.0-$$(git rev-list HEAD --count)-$$(git describe --dirty --always)" | sed 's/-/./2' | sed 's/-/./2')
```
jbw976: i validated that when there are no tags in a repo, the only thing generated from `git describe --dirty --always` is the commit short hash without the leading `g`, so this change does seem valid. I was worried it might create a duplicated leading `g`, but it does not seem to :)

PR #93, opened 2020-03-16
Allows build with Go 1.14

artursouza:
Signed-off-by: Artur Souza <artursouza.ms@outlook.com>

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -51,7 +51,7 @@ GO_LDFLAGS += -s -w
 endif
 
 # supported go versions
-GO_SUPPORTED_VERSIONS ?= 1.7|1.8|1.9|1.10|1.11|1.12|1.13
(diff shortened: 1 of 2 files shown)


makelib/golang.mk
```diff
@@ -51,7 +51,7 @@ GO_LDFLAGS += -s -w
 endif
 
 # supported go versions
-GO_SUPPORTED_VERSIONS ?= 1.7|1.8|1.9|1.10|1.11|1.12|1.13
+GO_SUPPORTED_VERSIONS ?= 1.7|1.8|1.9|1.10|1.11|1.12|1.13|1.14
```
jbw976: do we want to update the cross build container to use 1.14 during builds, similar to how we did that in https://github.com/upbound/build/commit/06c19d765514ee035bb6dbe582cd80c0313312ac#diff-a01ea346de0fbd4a8a8cfccb3b0d7793?

artursouza: Done.

PR #95, opened 2020-03-25
Make S3_BUCKET and publishing to it optional

soorena776:
WESAAS does not need publishing to an S3_BUCKET, so making it optional in  `output.mk`

Signed-off-by: Javad Taheri <taheri.javad@gmail.com>


jbw976:
Merging this PR since it's been approved by 2 folks

PR #96, opened 2020-04-07
Bump KIND version to v0.7.0

hasheddan:
This updates Kind to use k8s v1.17 and also adds some new features. There are some minor updates that will be required in https://github.com/crossplane/crossplane/blob/master/cluster/local/integration_tests.sh, which I am happy to make :+1: 

### DIFF
--- a/makelib/k8s_tools.mk
+++ b/makelib/k8s_tools.mk
@@ -16,7 +16,7 @@
 # Options
 
 # the version of kind to use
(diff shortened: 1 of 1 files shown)


hasheddan:
@jbw976 it actually appears that the integration tests won't actually break but they do log an error that I have shown the fix for here: https://github.com/crossplane/crossplane/pull/1405

PR #98, opened 2020-06-01
Bump cross container Go version to 1.14.4

hasheddan:
This updates the cross container go version to include a
workaround for the linux kernel bug described in this
issue https://github.com/golang/go/issues/37436


hasheddan:
I have tested this with https://github.com/crossplane/crossplane/pull/1581 and believe it is ready to be merged 👍

PR #110, opened 2020-09-01
Local dev: add script only component support

turkenh:
- istio support for local dev
- logging improvements with bold style echo_info

### DIFF
--- a/scripts/utils.sh
+++ b/scripts/utils.sh
@@ -1,21 +1,23 @@
 ################################# setup colors
 # setting up colors
-BLU='\033[0;34m'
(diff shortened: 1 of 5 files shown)


scripts/utils.sh
```diff
@@ -1,21 +1,23 @@
 ################################# setup colors
 # setting up colors
-BLU='\033[0;34m'
+BOLD='\033[10;1m'
```
lukeweber, line 3: Any reason bold vs blu? just to see it better?

turkenh: yes, just to see it better and differentiate from logs of other steps.

PR #119, opened 2020-09-30
Fixes for helm3

turkenh:
- helm3 no longer uses HELM_HOME env var: https://helm.sh/docs/faq/#xdg-base-directory-support (causing users directories to be used otherwise)
- we don't need helm init for localdev in case of helm3


lukeweber (approved):

ktintc (approved):
Thank you!

PR #122, opened 2020-10-20
Support Go 1.15

negz:
I've confirmed Crossplane core builds and runs by running:

 ```console
$ go version
go version go1.15.3 darwin/amd64

$ GO_SUPPORTED_VERSIONS=1.15 make distclean clean test build
```

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -51,7 +51,7 @@ GO_LDFLAGS += -s -w
 endif
 
 # supported go versions
-GO_SUPPORTED_VERSIONS ?= 1.7|1.8|1.9|1.10|1.11|1.12|1.13|1.14
+GO_SUPPORTED_VERSIONS ?= 1.7|1.8|1.9|1.10|1.11|1.12|1.13|1.14|1.15
 
 # set GOOS and GOARCH
 GOOS := $(OS)

### CODE REVIEW

File: makelib/golang.mk [hunk 1]
```diff
@@ -51,7 +51,7 @@ GO_LDFLAGS += -s -w
 endif
 
 # supported go versions
-GO_SUPPORTED_VERSIONS ?= 1.7|1.8|1.9|1.10|1.11|1.12|1.13|1.14
+GO_SUPPORTED_VERSIONS ?= 1.7|1.8|1.9|1.10|1.11|1.12|1.13|1.14|1.15
```
negz, line 54: Is this whitelist something we consider useful? Has it saved us in the past? I find that I mostly just validate new versions against one or two of our projects, then submit a PR to bump it here. It increases slightly annoying friction to upgrade to new versions (i.e. update this, then update 5-10 submodules).

I wonder if a better pattern would be to require each project to declare their supported versions, or even to switch to a blacklist of versions that are known not to work for whatever reason?

hasheddan, line 54: Should we update to actually build with go 1.15 also? I think the purpose of this list is to essentially set a maximum supported version. So we usually keep the maximum version in lockstep with the version that we build with so that consumers of the submodule are not surprised by something building locally but not in CI. And in that case, the build submodule bump is needed across repos anyway to actually build with the new version we download in the cross container.

muvaf: > It introduces slightly annoying friction to upgrade to new versions (i.e. update this, then update 5-10 submodules).

I agree. I think when this list was written, there wasn't go modules. Today, I think we should be able to let the project compile with the go version it declares in its [`go.mod`](https://github.com/crossplane/crossplane/blob/c683189/go.mod#L3)

> The go directive in a go.mod file now indicates the version of the language used by the files within that module. It will be set to the current release (go 1.12) if no existing version is present. If the go directive for a module specifies a version newer than the toolchain in use, the go command will attempt to build the packages regardless, and will note the mismatch only if that build fails. 

From https://golang.org/doc/go1.12#modules

PR #125, opened 2020-11-06
make go covermod configurable

wonderflow:
Signed-off-by: 天元 <jianbo.sjb@alibaba-inc.com>

`go test -race` need CGO_ENABLED=1.

Tested with my personal repo: https://github.com/wonderflow/test-build

1. add build as submodule with this updated PR
2. run make test

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -130,6 +130,7 @@ GO_COMMON_FLAGS = $(GO_BUILDFLAGS) -tags '$(GO_TAGS)'
 GO_STATIC_FLAGS = $(GO_COMMON_FLAGS) $(GO_PKG_STATIC_FLAGS) -installsuffix static  -ldflags '$(GO_LDFLAGS)'
 GO_GENERATE_FLAGS = $(GO_BUILDFLAGS) -tags 'generate $(GO_TAGS)'
 GO_COVER_MODE ?= count
+CGO_ENABLED ?= 0
 
 export GO111MODULE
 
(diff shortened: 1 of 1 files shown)


wonderflow:
ping @negz @hasheddan @jbw976  , please kindly review, thanks!

hasheddan (approved):
LGTM

jbw976 (commented):
Thanks for this contribution @wonderflow! Can you please update the PR description with some details about how this has been tested? Thank you!

wonderflow:
@jbw976  Thanks , I've tested it.

makelib/golang.mk
```diff
@@ -129,6 +129,8 @@ endif
 GO_COMMON_FLAGS = $(GO_BUILDFLAGS) -tags '$(GO_TAGS)'
 GO_STATIC_FLAGS = $(GO_COMMON_FLAGS) $(GO_PKG_STATIC_FLAGS) -installsuffix static  -ldflags '$(GO_LDFLAGS)'
 GO_GENERATE_FLAGS = $(GO_BUILDFLAGS) -tags 'generate $(GO_TAGS)'
+GO_COVER_MODE ?= count
```
jbw976, line 132: i would recommend putting this up near the top of this file, around line 31 that says `# Optional build flags passed to go tools`.  That will help people more easily discover the set of config options that they can set.

nit: consider calling `CGO_ENABLED` -> `GO_CGO_ENABLED` instead, for someone doing a search for `^GO_` can find it along with the other options that are named similarly.

makelib/golang.mk
```diff
@@ -171,7 +173,7 @@ ifeq ($(GO_NOCOV),true)
 else
 	@mkdir -p $(GO_TEST_OUTPUT)
 	@CGO_ENABLED=0 $(GOHOST) test -i -cover $(GO_STATIC_FLAGS) $(GO_PACKAGES) || $(FAIL)
```
jbw976, line 175: is it odd that this `test` command will always run with `CGO_ENABLED=0`, but the `test` command on the next line will use the `CGO_ENABLED` variable?

makelib/golang.mk
wonderflow: Thanks, I followed the advice.

makelib/golang.mk
wonderflow: Sure, I could change this line. But I noticed the `@CGO_ENABLED=0` was hard coded lots of places, I don't want this PR make widely changes

wonderflow:
@jbw976  All fixed and again tested in  https://github.com/wonderflow/test-build, Thanks!

jbw976 (approved):
looks good @wonderflow! thank you for incorporating the feedback and for driving this improvement! 💪

PR #126, opened 2020-11-09
Remove lint from non-relevant places

khos2ow:
Remove lint jobs and targets from processes and targets which they don't necessarily need to run lint. Notable changes:

- `make build`: `make build.check` was removed from it
- `make helm.check`: new target added for linting helm chart
- `make go.vendor.tidy`: new target for `go mod tidy` which will be used in upstream projects' targets (e.g. `make reviewable`).

This PR blocks https://github.com/crossplane/crossplane/pull/1950 and should be reviewed as a pair.

### DIFF
--- a/makelib/common.mk
+++ b/makelib/common.mk
@@ -300,7 +300,6 @@ do.build.artifacts: $(foreach p,$(PLATFORMS), do.build.artifacts.$(p))
 # build for all platforms
 build.all:
 	@$(MAKE) build.init
-	@$(MAKE) build.check
 	@$(MAKE) build.code
 	@$(MAKE) do.build.platform
 	@$(MAKE) build.artifacts
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -232,6 +232,11 @@ go.vendor:
 	@$(GOHOST) mod vendor || $(FAIL)
 	@$(OK) go mod vendor
 
+go.vendor.tidy:
+	@$(INFO) go mod tidy
+	@$(GOHOST) mod tidy || $(FAIL)
+	@$(OK) go mod tidy
+
 else
 
 go.vendor.lite: $(DEP)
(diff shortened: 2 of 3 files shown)


hasheddan (commented):
@khos2ow is the title for this PR related to the intended behavior? It looks like we just call `go mod tidy` directly in `c/c`, so it isn't immediately apparent to me how this facilitates removing `lint`

khos2ow:
@hasheddan No, I should've converted it to draft a while ago. This is me attempting to move linting only to relevant CI jobs and targets. I was about to update this PR and consecutively open a corresponding PR in c/c.

khos2ow:
@negz @hasheddan this should be good for review now.

makelib/common.mk
```diff
@@ -300,7 +300,6 @@ do.build.artifacts: $(foreach p,$(PLATFORMS), do.build.artifacts.$(p))
 # build for all platforms
 build.all:
 	@$(MAKE) build.init
-	@$(MAKE) build.check
```
khos2ow, line 303: The most notable change here is that this line is removed. As I mentioned before this is really debatable. It's quite nice to have it executed as part of the build process, but at the same time the CI is in place and does that, separately on different job.

negz: One thing to note here is that this build submodule is used by several Upbound internal projects, so these changes could affect more than just Crossplane. :( I think we eventually should fork the build library under the Crossplane org so that folks without visibility into Upbound's private repos can make changes safely.

This was mentioned in https://github.com/crossplane/crossplane/issues/1583, which I accidentally closed while only half done.

khos2ow: That's a really valid point! Maybe as an interim workaround (before forking this under `crossplane` org) an environment variable can be added to disable this.

```makefile
RUN_BUILD_CHECK ?=true
build.all:
	@$(MAKE) build.init
ifeq ($(RUN_BUILD_CHECK),"true")
	@$(MAKE) build.check
endif
```

PR #129, opened 2020-12-23
Make helm3 version configurable

zzxwill:
Due to the announcement form Helm https://helm.sh/blog/new-location-stable-incubator-charts/,
the new location for helm chart repository has been upgraded to Github Pages. We MUST update the
repositories we use before November 13, 2020. Helm v2.17.0 and v3.4.0 have been released to
help use the new location.

Signed-off-by: zzxwill <zzxwill@gmail.com>

### DIFF
--- a/makelib/k8s_tools.mk
+++ b/makelib/k8s_tools.mk
@@ -38,7 +38,7 @@ KUSTOMIZE := $(TOOLS_HOST_DIR)/kustomize-$(KUSTOMIZE_VERSION)
 
 # the version of helm 3 to use
 USE_HELM3 ?= false
-HELM3_VERSION := v3.2.4
+HELM3_VERSION ?= v3.4.0
 HELM3 := $(TOOLS_HOST_DIR)/helm-$(HELM3_VERSION)
 
 # If we enable HELM3 we alias HELM to be HELM3
--- a/makelib/k8s_tools.mk
+++ b/makelib/k8s_tools.mk
@@ -38,7 +38,7 @@ KUSTOMIZE := $(TOOLS_HOST_DIR)/kustomize-$(KUSTOMIZE_VERSION)
 
 # the version of helm 3 to use
 USE_HELM3 ?= false
-HELM3_VERSION ?= v3.4.0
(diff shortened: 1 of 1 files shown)


wonderflow:
@negz @hasheddan  Please help review, our CI is blocked due to this. https://github.com/crossplane/oam-kubernetes-runtime/pull/316

makelib/k8s_tools.mk
```diff
@@ -38,15 +38,15 @@ KUSTOMIZE := $(TOOLS_HOST_DIR)/kustomize-$(KUSTOMIZE_VERSION)
 
 # the version of helm 3 to use
 USE_HELM3 ?= false
-HELM3_VERSION := v3.2.4
+HELM3_VERSION := v3.4.0
```
wonderflow, line 41:
```suggestion
HELM3_VERSION ?= v3.2.4
```

makelib/k8s_tools.mk
```diff
@@ -38,7 +38,7 @@ KUSTOMIZE := $(TOOLS_HOST_DIR)/kustomize-$(KUSTOMIZE_VERSION)
 
 # the version of helm 3 to use
 USE_HELM3 ?= false
-HELM3_VERSION := v3.2.4
+HELM3_VERSION ?= v3.4.0
```
jbw976, line 41: yes this is definitely better to allow this value to already be set to the callers desired version, thanks for making it.

what do you think about setting the default to v3.4.2, the latest patch version for the v3.4 release?
https://github.com/helm/helm/releases/tag/v3.4.2

zzxwill: +1. Submitted the other commit. Do you think it's necessary to set helm version 2 to the version above v2.17.0 too, the latest one?

jbw976: thanks! no i don't think it's necessary to also update version 2 at the same time. this PR is focused (and has a distinct need for updating the version of) on version 3 so we can just keep that focus.

jbw976 (approved):
thank you for making this update, future folks will be grateful that they can set their own version of helm3 now too :)

PR #130, opened 2021-01-05
Update default base image to alpine:3.12 and add support for armv6

hasheddan:
`alpine:3.7` is no longer supported, so we update to `alpine:3.12` as default, which also supports `armv6`. 32-bit arm golang builds default to `armv6` and we don't currently allow for overriding `GOARM` in `build` submodule.

### DIFF
--- a/makelib/image.mk
+++ b/makelib/image.mk
@@ -35,6 +35,8 @@ ifeq ($(ARCH),amd64)
 OSBASEIMAGE = $(OSBASE)
 else ifeq ($(ARCH),arm64)
 OSBASEIMAGE = arm64v8/$(OSBASE)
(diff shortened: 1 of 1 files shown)


hasheddan:
Example usage: https://github.com/crossplane/crossplane/pull/2061/checks?check_run_id=1651804036

Knappek:
@hasheddan do we also need to add `linux_arm` to the [Dockerfile](https://github.com/upbound/build/blob/master/cross/Dockerfile#L58)?

hasheddan:
@Knappek good catch! I am assuming that it is working okay currently because it isn't a requirement that we have a pre-compiled standard lib, but need to double check :+1:

PR #137, opened 2021-02-12
Allow build with mingw64

MisterMX:
The make process should allow build on windows using MINGW64 which provides a bash running on Windows (without WSL).

One issue is that MINGW64's built-in `uname` returns something like `MINGW64_NT-10.0-19041` in opposite to `linux` or `darwin` on other systems.

### Description of changes

1. In `common.mk` check if `HOSTOS` starts with `mingw64_nt` and if yes, set `HOSTOS := windows`
2. Add `windows` to allowed platforms
3. Continue as before

### How has the code been tested

Tested with https://github.com/crossplane/provider-aws using Make `4.3` and Golang `1.15.7`.


MisterMX:
Closing this as it is not relevant for me anymore.

I recommend switching to *nix like OS or use a VM if you are stuck with Windows.

PR #140, opened 2021-03-11
Build with Go 1.16.1, and remove the whitelist of supported versions

negz:
Closes https://github.com/upbound/build/pull/122

Per https://github.com/upbound/build/pull/122#discussion_r508284675, most (all?) of our projects now use Go modules, which allow a module to declare the version of Go it targets.

It could be that the purpose of this variable is to gate the versions of Go that `upbound/build` as a library of Makefiles supports, rather than the versions of Go that _code built with_ `upbound/build` supports. If this is the case, I'd argue that during the the two years I've been using this library we've yet to encounter a new version of Go that has required anything other than a bump to this whitelist and that we never test that the older versions of Go still work. I suspect it would be more productive to pay the price of a newer version of Go breaking the library if and when that becomes an issue than it currently is to bump the var every time a new version of Go is released.

Edit: I've now tested this by:

* Confirming I can build `crossplane/crossplane` with this PR, and Go 1.16.1.
* Confirming I can `make img.build && docker run <crosscontainer> go version` for the Cross container.

### DIFF
--- a/cross/Dockerfile
+++ b/cross/Dockerfile
@@ -45,8 +45,8 @@ RUN curl -fsSL "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-2.1.21.zip"
 
 # ------------------------------------------------------------------------------------------------
 # Go support
-RUN GO_VERSION=1.14.4 && \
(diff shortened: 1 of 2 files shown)


jbw976 (approved):
I think your reasoning is sound here @negz in that the enforcement of this check has caused more headaches than it has saved us from.

PR #141, opened 2021-03-15
Option to disable helm strict linting

turkenh:
This PR adds an option to disable strict helm linting which fails the build with warnings. Ideally, we should fix warnings instead of disabling strict linting, however, sometimes this is not possible (or does not make sense) like an incorrect warning on the helm side. 

Currently, [this issue](https://github.com/helm/helm/issues/9197) on helm side prevents use prevents us using helm 3 with Crossplane chart. With this option, we will be able to (temporarily) disable strict linting.

This PR also updates helm version to latest (v3.5.3) which includes PRs fixing[ categorization of linting issues ](https://github.com/helm/helm/issues/8596)as warnings instead of errors.

### DIFF
--- a/scripts/localdev-deploy-component.sh
+++ b/scripts/localdev-deploy-component.sh
@@ -66,9 +66,9 @@ if [ "${LOCALDEV_LOCAL_BUILD}" == "true" ] && containsElement "${HELM_CHART_NAME
   [ -f "${HELM_CHART_REF}" ] || echo_error "Local chart ${HELM_CHART_REF} not found. Did you run \"make build\" ? "
 
   # If local build, tag "required" local images, so that they can be load into kind cluster at a later step.
-  for r in "${registries_arr[@]}"; do
-    for i in "${images_arr[@]}"; do
-      for a in "${image_archs_arr[@]}"; do
+  for r in ${registries_arr[@]+"${registries_arr[@]}"}; do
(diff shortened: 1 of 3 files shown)


scripts/localdev-deploy-component.sh
```diff
@@ -66,9 +66,9 @@ if [ "${LOCALDEV_LOCAL_BUILD}" == "true" ] && containsElement "${HELM_CHART_NAME
   [ -f "${HELM_CHART_REF}" ] || echo_error "Local chart ${HELM_CHART_REF} not found. Did you run \"make build\" ? "
 
   # If local build, tag "required" local images, so that they can be load into kind cluster at a later step.
-  for r in "${registries_arr[@]}"; do
-    for i in "${images_arr[@]}"; do
-      for a in "${image_archs_arr[@]}"; do
+  for r in ${registries_arr[@]+"${registries_arr[@]}"}; do
+    for i in ${images_arr[@]+"${images_arr[@]}"}; do
+      for a in ${image_archs_arr[@]+"${image_archs_arr[@]}"}; do
```
hasheddan, line 71: TIL I learned this is [the only consistently safe expansion pattern](https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u).

hasheddan (approved):
LGTM!

PR #143, opened 2021-03-22
upgrade tini to v0.19.0

smcavallo:
See - https://github.com/krallin/tini/releases/tag/v0.19.0
v0.16.1 -> v0.17.0 -> v0.18.0 -> v0.19.0
none of the above contain breaking changes
Signed-off-by: smcavallo <smcavallo@hotmail.com>


hasheddan (commented):
@smcavallo i wonder if we can remove the Tini setup and use what is built into Docker now? https://github.com/krallin/tini

PR #144, opened 2021-03-22
build dockerfile using newer ubuntu release

smcavallo:
Build environment should use newer ubuntu release
Xenial is EOL on 4/2021 - https://help.ubuntu.com/community/EOL#Ubuntu_16.04_Xenial_Xerus
Was able to build this dockerfile locally without issue.
Ex. `docker build --build-arg TINI_VERSION=v0.19.0 --build-arg ARCH=amd64 .`
Signed-off-by: smcavallo <smcavallo@hotmail.com>

### DIFF
--- a/cross/Dockerfile
+++ b/cross/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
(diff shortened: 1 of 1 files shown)


hasheddan (commented):
@smcavallo thanks! Can you make sure that cross builds are working correctly with something like `./build/run make build.all` In the Crossplane repo?

muvaf:
Why not use `20.04`, the latest LTS?

PR #151, opened 2021-05-05
Use SAFEHOSTPLATFORM for olm-bundle URL

muvaf:


makelib/common.mk
```diff
@@ -132,7 +132,6 @@ endif
 # Standardize Host Platform variables
 HOST_PLATFORM := $(HOSTOS)_$(HOSTARCH)
 SAFEHOSTPLATFORM := $(HOSTOS)-$(SAFEHOSTARCH)
```
muvaf, line 134: @AaronME is there a specific reason both `SAFEHOST_PLATFORM` and `SAFEHOSTPLATFORM` exists? And would you have a preference between them?

makelib/k8s_tools.mk
```diff
@@ -97,34 +97,34 @@ $(KUBECTL):
 $(KUSTOMIZE):
 	@$(INFO) installing kustomize $(KUSTOMIZE_VERSION)
 	@mkdir -p $(TOOLS_HOST_DIR)/tmp-kustomize
-	@curl -fsSL https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/$(KUSTOMIZE_VERSION)/kustomize_$(KUSTOMIZE_VERSION)_$(SAFEHOST_PLATFORM).tar.gz | tar -xz -C $(TOOLS_HOST_DIR)/tmp-kustomize
+	@curl -fsSL https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/$(KUSTOMIZE_VERSION)/kustomize_$(KUSTOMIZE_VERSION)_$(SAFEHOSTPLATFORM).tar.gz | tar -xz -C $(TOOLS_HOST_DIR)/tmp-kustomize
```
hasheddan, line 100: Will this break since these two variables are not interchangeable?

makelib/common.mk
hasheddan: @muvaf note that one use `-` and one uses `_`

makelib/k8s_tools.mk
AaronME: Based on https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh the kustomize install script is setting the version to "${opsys}_${arch}" -- so SAFEHOST_PLATFORM should be correct.

makelib/common.mk
muvaf: Wow I did not notice that, reverted the removal. Thanks folks!

turkenh (approved):

PR #155, opened 2021-05-21
Fix version for two hypens in tag

turkenh:
When the tag contains two hyphens, we end up generating versions not sortable as semver. See https://github.com/upbound/universal-crossplane/issues/109 for more details.

This PR fixes that by converting `-` to `.` in that case just like we do other hyphens in `git describe` output.

### DIFF
--- a/makelib/common.mk
+++ b/makelib/common.mk
@@ -210,10 +210,10 @@ ifeq ($(origin VERSION), undefined)
 # check if there are any existing `git tag` values
 ifeq ($(shell git tag),)
 # no tags found - default to initial tag `v0.0.0`
-VERSION := $(shell echo "v0.0.0-$$(git rev-list HEAD --count)-g$$(git describe --dirty --always)" | sed 's/-/./2' | sed 's/-/./2')
+VERSION := $(shell echo "v0.0.0-$$(git rev-list HEAD --count)-g$$(git describe --dirty --always)" | sed 's/-/./2' | sed 's/-/./2' | sed 's/-/./2')
 else
 # use tags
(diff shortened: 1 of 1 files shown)


makelib/common.mk
```diff
@@ -210,10 +210,10 @@ ifeq ($(origin VERSION), undefined)
 # check if there are any existing `git tag` values
 ifeq ($(shell git tag),)
 # no tags found - default to initial tag `v0.0.0`
-VERSION := $(shell echo "v0.0.0-$$(git rev-list HEAD --count)-g$$(git describe --dirty --always)" | sed 's/-/./2' | sed 's/-/./2')
+VERSION := $(shell echo "v0.0.0-$$(git rev-list HEAD --count)-g$$(git describe --dirty --always)" | sed 's/-/./2' | sed 's/-/./2' | sed 's/-/./2')
 else
 # use tags
-VERSION := $(shell git describe --dirty --always --tags | sed 's/-/./2' | sed 's/-/./2' )
+VERSION := $(shell git describe --dirty --always --tags | sed 's/-/./2' | sed 's/-/./2' | sed 's/-/./2' )
```
muvaf, line 216: nitpick: I had to check what `2` in `s/-/./2` [means](https://www.gnu.org/software/sed/manual/sed.html#sed-script-overview) and apparently it says replace only the 2nd instance, which saves us from replacing the first `-` in for example `v1.3.0-rc.0-42-g6622b161`. A comment explaining that could be helpful for future readers.

muvaf (approved):
LGTM!

makelib/common.mk
turkenh: yes, this is correct.

PR #156, opened 2021-05-27
Add a common `make reviewable` target

negz:

We've been using this target in Crossplane projects for some time. The idea is that's it's a one-shot folks can run to validate that their branch is not likely to fail CI when they open a branch, by ensuring codegen has run, and that any linters and tests pass.

I have:

- [ ] Added `backport release-x.y` labels to auto-backport this PR.

I've validated that this works with crossplane/crossplane (with a few tweaks to its own `Makefile`).

### DIFF
--- a/makelib/common.mk
+++ b/makelib/common.mk
@@ -418,7 +418,13 @@ reviewable:
 	@$(MAKE) lint
 	@$(MAKE) test
 
-.PHONY: publish.init publish.artifacts publish promote.init promote.artifacts promote tag generate reviewable
+# ensure reviewable target doesn't create a diff
+check-diff: reviewable
+	@$(INFO) checking that branch is clean
+	@test -z "$(shell git status --porcelain)"" || $(FAIL)
+	@$(OK) branch is clean
+
+.PHONY: publish.init publish.artifacts publish promote.init promote.artifacts promote tag generate reviewable check-diff
 
 # ====================================================================================
 # Help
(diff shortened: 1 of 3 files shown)


makelib/common.mk
```diff
@@ -409,7 +409,29 @@ promote:
...
+# prepare for code review
+reviewable:
+	@$(MAKE) generate
+	@$(MAKE) lint
+	@$(MAKE) test
+
+# ensure reviewable target doesn't create a diff
+check-diff: reviewable
+	@$(INFO) checking that branch is clean
+	@test -z "$(shell git status --porcelain)"" || $(FAIL)
```
hasheddan, line 431:
```suggestion
	@test -z "$(shell git status --porcelain)" || $(FAIL)
```

makelib/common.mk
```diff
@@ -426,6 +448,9 @@ Common Targets:
     help         Show this help info.
     test         Runs unit tests.
     e2e          Runs end-to-end integration tests.
+	generate     Run code generation.
+	reviewable   Validate that a PR is ready for review.
+	check-diff   Ensure the reviewable target doesn't create a git diff.
```
hasheddan: Are these slightly offset or is that just the rendering?

hasheddan: ah looks like you fixed as I posted 👍

makelib/common.mk
negz: Fixed! I also refactored `check-diff` a bit so it won't run the linter or tests needlessly.

hasheddan (approved):

PR #159, opened 2021-06-11
go: do not run go.lint for every build

muvaf:

`go.lint` takes quite a while for big repositories like provider-aws but it's actually not relevant for building. The check is done by other mechanisms like `make reviewable` when necessary. This PR removes the `go.lint` from the `build` target.

I have:

Manually with provider-aws

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -281,7 +281,6 @@ go.generate:
 # Common Targets
 
 build.init: go.init
-build.check: go.lint
 build.code.platform: go.build
 clean: go.clean
 distclean: go.distclean


zzxwill (approved):
/lgtm

hasheddan (commented):
I wasn't sure about this initially, but I think after framing it in the following manner it makes sense:
- The purpose of building is to take valid source (i.e. source that the compiler will accept) and produce a binary.
- The purpose of linting is to make sure that we do not check in source that does not adhere to our standards.

IMO if we are linting somewhere on a blocking PR review then that is sufficient as any source that is checked in and valid should be able to be built. Therefore, I would be in favor of this change. However, I do know at @khos2ow opened a similar PR (https://github.com/upbound/build/pull/126) (with slightly larger scope) suggesting this change, so if we are in favor I think we should drive that one to being mergeable.

negz:
I feel okay about this change.

It does slightly increase our risk, in that it's now effectively up to our CI system to enforce that the linter passes. i.e. If someone setup CI without a required `make reviewable` step (or more likely the golangci-lint action we use as part of the ci.yaml GitHub workflow) it will be possible for them to build and ship artifacts that did not pass the linter. I think we can tolerate that risk though; we'll just need to make sure all repos enforce the linter check.

Do we have an understanding of whether this change makes sense for our internal Upbound projects that build using CircleCI? I suspect those repos are relying on the dependency we're removing in this PR to trigger the linter.

> I do know at @khos2ow opened a similar PR (#126) (with slightly larger scope) suggesting this change, so if we are in favor I think we should drive that one to being mergeable.

Could we work with @khos2ow to make a handful of smaller changes?

khos2ow:
> Do we have an understanding of whether this change makes sense for our internal Upbound projects that build using CircleCI?

Maybe now could be a good time to proceed with forking this repo dedicated to Crossplane organization? Frankly this will solely be for Upbound folks to decide.

muvaf:
> Do we have an understanding of whether this change makes sense for our internal Upbound projects that build using CircleCI? I suspect those repos are relying on the dependency we're removing in this PR to trigger the linter.

I'd say that not running linter during build could be considered a bug fix in general per the reasons @hasheddan listed, so it'd make sense for them as well IMO. Seems like it's more of making sure those actions still run linter in one way or another. Since the build submodule update is done manually via a PR, I'd rely on the maintainers of those repos to decide what to do when they update the submodule.

negz:
> I'd rely on the maintainers of those repos to decide what to do when they update the submodule.

I think in reality we'll need more coordination than that - AFAIK the CircleCI setups we have at the moment are unmaintained so if we proceed with this I'd like you to either take updating those or find folks willing to do so (maybe @AaronME). I'm all for speeding up builds, but not at the expense of removing CI coverage elsewhere without a plan to address it.

negz (approved):

PR #161, opened 2021-07-29
Adds support for darwin amd64 and arm64

tnthornton:
Signed-off-by: Taylor Thornton <taylor@upbound.io>

The current build submodule makes an effort to not use darwin_arm64. Unfortunately, being on an Apple M1 MBP, that makes it so certain builds don't quite work.

In addition to the script updates, I upgraded kind as the earlier version defined would result in:
```
ERROR: failed to create cluster: failed to init node with kubeadm: command "docker exec --privileged build-07a4805a-inttests-control-plane kubeadm init --skip-phases=preflight --config=/kind/kubeadm.conf --skip-token-print --v=6" failed with error: exit status 139
Command Output: SIGSEGV: segmentation violation
PC=0x0 m=0 sigcode=0

goroutine 1 [running]:
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
```
when running `make e2e` on crossplane/crossplane.

I have:

These changes are coming in from a local branch I was using to get crossplane tests to work, specifically:
```
make
make test
make e2e
```

### DIFF
--- a/makelib/common.mk
+++ b/makelib/common.mk
@@ -105,13 +105,6 @@ endif
 # Set the host's arch.
 HOSTARCH := $(shell uname -m)
 
-# Set safe architectures for supported OSes (darwin and linux).
-# Apple Silicon binaries are not widely available yet
-ifeq ($(HOSTOS),darwin)
-SAFEHOSTARCH := amd64
-TARGETARCH := $(HOSTARCH)
-endif
-
 # If SAFEHOSTARCH and TARGETARCH have not been defined yet, use HOST
 ifeq ($(origin SAFEHOSTARCH),undefined)
 SAFEHOSTARCH := $(HOSTARCH)
--- a/makelib/common.mk
+++ b/makelib/common.mk
@@ -105,6 +105,13 @@ endif
 # Set the host's arch.
 HOSTARCH := $(shell uname -m)
 
+# Set safe architectures for supported OSes (darwin and linux).
+# Apple Silicon binaries are not widely available yet
+ifeq ($(HOSTOS),darwin)
+SAFEHOSTARCH := amd64
+TARGETARCH := $(HOSTARCH)
+endif
+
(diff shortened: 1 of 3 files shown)


makelib/common.mk
```diff
@@ -105,13 +105,6 @@ endif
 # Set the host's arch.
 HOSTARCH := $(shell uname -m)
 
-# Set safe architectures for supported OSes (darwin and linux).
-# Apple Silicon binaries are not widely available yet
-ifeq ($(HOSTOS),darwin)
-SAFEHOSTARCH := amd64
```
AaronME, line 111: SAFEHOSTARCH has been necessary because of dependencies in https://github.com/upbound/build/blob/master/makelib/golang.mk and https://github.com/upbound/build/blob/master/makelib/k8s_tools.mk

At the beginning of the year several tools did not offer arm64 binaries. So we would fall back to amd64, which the m1 can run via rosetta.

tnthornton: Gotcha. I removed it here due to some image tagging that's happening in crossplane. I can find a different way of handling it in the scripts over there if need be.

makelib/common.mk
```diff
@@ -105,13 +105,6 @@ endif
 # Set the host's arch.
 HOSTARCH := $(shell uname -m)
 
-# Set safe architectures for supported OSes (darwin and linux).
-# Apple Silicon binaries are not widely available yet
```
hasheddan, line 109: @tnthornton I think this will at least break OLM bundle (https://github.com/upbound/olm-bundle/blob/f90948c2a3fb344f6b1d13b346b9ca1b62c89c9b/Makefile#L4) and potentially a few of the other tools. I wonder if we can either:
1. Update to compatible versions of each tool.
2. Say that you should set SAFEHOSTARCH manually if you are on Apple silicon and want to have a good time.

AaronME (approved):

PR #162, opened 2021-08-05
Add dep update step so that charts with subcharts don't fail linting

tnthornton:

In Helm3, the linting rules have changed slightly. Now if you have a dependency defined in your Chart.yaml, but don't have a corresponding subchart directory, linting will fail. This change adds a step to ensure helm has the full view of the dependencies when it attempts linting.

I have:

- [X] Run `make reviewable` to ensure this PR is ready for review.

Locally on the enterprise-server repo.

### DIFF
--- a/makelib/helm.mk
+++ b/makelib/helm.mk
@@ -92,6 +92,7 @@ helm.prepare: helm.prepare.$(1)
 
 helm.lint.$(1): $(HELM_HOME) helm.prepare.$(1)
 	@rm -rf $(abspath $(HELM_CHARTS_DIR)/$(1)/charts)
+	@$(HELM) dependency update $(abspath $(HELM_CHARTS_DIR)/$(1))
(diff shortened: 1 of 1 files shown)


makelib/helm.mk
```diff
@@ -92,6 +92,7 @@ helm.prepare: helm.prepare.$(1)
 
 helm.lint.$(1): $(HELM_HOME) helm.prepare.$(1)
 	@rm -rf $(abspath $(HELM_CHARTS_DIR)/$(1)/charts)
+	@$(HELM) dependency update $(abspath $(HELM_CHARTS_DIR)/$(1))
```
hasheddan, line 95: Just noting from our conversation that the reason that we are not gating this on Helm 3 is because you should theoretically want to lint with your dependencies present anyway as they will eventually get pulled in during build (`helm.dep` is dependency of `build.check`).

hasheddan (approved):
Thanks @tnthornton !

PR #163, opened 2021-08-12
Add helm repo add step if internal_museum_url is defined

tnthornton:
This change adds an additional init step that adds our internal helm chart museum as a repo. We do this to enable authenticated requests to it in the event the `INTERNAL_MUSEUM_URL` is defined, as is the case with enterprise-server.

I have:

Primarily tested in the enterprise-server repo. Both disabled, i.e. not specifying `INTERNAL_MUSEUM_URL` and verifying that issuing `make` did not result in error as well as specifying `INTERNAL_MUSEUM_URL` and verifying that the `helm repo add` command was executed and the desired `pull` behavior was seen when the accompanying _USER and _PASS env vars we're also configured.

### DIFF
--- a/makelib/helm.mk
+++ b/makelib/helm.mk
@@ -84,6 +84,11 @@ $(HELM_OUTPUT_DIR)/$(1)-$(HELM_CHART_VERSION).tgz: $(HELM_HOME) $(HELM_OUTPUT_DI
 	fi
 	@$(OK) helm package $(1) $(HELM_CHART_VERSION)
 
+helm.repo.add:
+ifdef INTERNAL_MUSEUM_URL
+	@$(HELM) repo add $(INTERNAL_MUSEUM_NAME) $(INTERNAL_MUSEUM_URL) --username $(INTERNAL_MUSEUM_USER) --password $(INTERNAL_MUSEUM_PASS)
+endif
+
 helm.prepare.$(1): $(HELM_HOME)
 	@cp -f $(HELM_CHARTS_DIR)/$(1)/values.yaml.tmpl $(HELM_CHARTS_DIR)/$(1)/values.yaml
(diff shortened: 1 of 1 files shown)


makelib/helm.mk
```diff
@@ -84,6 +84,11 @@ $(HELM_OUTPUT_DIR)/$(1)-$(HELM_CHART_VERSION).tgz: $(HELM_HOME) $(HELM_OUTPUT_DI
 	fi
 	@$(OK) helm package $(1) $(HELM_CHART_VERSION)
 
+helm.repo.add:
+ifdef INTERNAL_MUSEUM_URL
+	@$(HELM) repo add $(INTERNAL_MUSEUM_NAME) $(INTERNAL_MUSEUM_URL) --username $(INTERNAL_MUSEUM_USER) --password $(INTERNAL_MUSEUM_PASS)
+endif
+
```
hasheddan, line 91: @tnthornton how much effort would if be to make this more generic such that you can define a list of dependency repos (and optionally credential variables) then we add them all? Another option would be that if you just want to do this for one repo then you can just extend `build.init` target in its specific `Makefile`. That might be easier for now. See example in: https://github.com/upbound/up/blob/b812d4c29487e47ee73db6fe7778e7c19ec8843a/Makefile#L54

tnthornton: >@tnthornton how much effort would if be to make this more generic such that you can define a list of dependency repos (and optionally credential variables) then we add them all?

Oh interesting. Let me play around with that and see.

>Another option would be that if you just want to do this for one repo then you can just extend build.init target in its specific Makefile. That might be easier for now. See example in: https://github.com/upbound/up/blob/b812d4c29487e47ee73db6fe7778e7c19ec8843a/Makefile#L54

I'll keep this in mind if I can't get the other suggestion to work.

PR #164, opened 2021-08-17
Update helm 2 download location

hasheddan:

As part of dropping support for Helm 2, the location for all published
helm versions has been moved to get.helm.sh. The currently used link is
now failing.

See https://github.com/helm/helm/issues/10005 and https://helm.sh/blog/get-helm-sh/

Also see example failure in https://github.com/crossplane/crossplane/pull/2484/checks?check_run_id=3350283419

I have:

This has not been individually tested, but it matches the same steps we currently use for Helm 3 (see a few lines below in same file). We could probably combine these two, but I am keeping them separate for now to preserve the required explicit declaration to use Helm 3.


AaronME (approved):
Confirmed that the following downloads successfully:

https://get.helm.sh/helm-v2.16.7-linux-amd64.tar.gz

LGTM!

PR #167, opened 2021-08-28
feat(check-diff): generate check-diff output

haarchri:
Signed-off-by: Christopher Haar <chhaar30@googlemail.com>

### Description of your changes

show check-diff modified output for specific files

old:
```
10:44:14 [ .. ] checking that branch is clean
10:44:14 [FAIL]
make: *** [check-diff] Error 1
```

new:
```
10:50:25 [ .. ] checking that branch is clean
 M build
 M package/crds/xxx.aws.crossplane.io_xyz.yaml
 M package/crds/xxx.aws.crossplane.io_abc.yaml
10:50:25 [FAIL]
```

I have:

- [x] Read and followed Upbound's [contribution process](https://git.io/fj2m9).
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR, as appropriate.

### How has this code been tested
via crossplane/provider-aws

PR #168, opened 2021-08-29
feat(check-diff): generate check-diff output

haarchri:
Signed-off-by: Christopher Haar <chhaar30@googlemail.com>

@hasheddan  missed few stuff in my branch we merged with #167 - one commit was missing  - now the feature is working - sorry for that

```
13:43:20 [ .. ] checking that branch is clean
13:43:20 [ OK ] branch is clean
```

```
13:47:06 [ .. ] checking that branch is clean
M Makefile
13:47:06 [FAIL]
make: *** [check-diff] Error 1
```

Fixes #

I have:

crossplane/provider-aws

### DIFF
--- a/makelib/common.mk
+++ b/makelib/common.mk
@@ -428,8 +428,7 @@ reviewable:
 # ensure generate target doesn't create a diff
 check-diff: generate
 	@$(INFO) checking that branch is clean
-	@git status --porcelain | grep . && $(FAIL)
-	@$(OK) branch is clean
+	@if git status --porcelain | grep . ; then $(FAIL); else $(OK) branch is clean; fi
 
 .PHONY: publish.init publish.artifacts publish promote.init promote.artifacts promote tag generate reviewable check-diff
 


makelib/common.mk
```diff
@@ -428,7 +428,12 @@ reviewable:
 # ensure generate target doesn't create a diff
 check-diff: generate
 	@$(INFO) checking that branch is clean
-	@git status --porcelain | grep . && $(FAIL)
```
hasheddan, line 431: @haarchri what was causing this strategy not to work? I had tested it out locally and it seemed to be effective.

haarchri: at the moment it fails everytime in clean and dirty state - because the check is missing ...

hasheddan: @haarchri the issue here is that we are always evaluating to `1` because we are either returning `1` from `git status --porcelain | grep .` (when the branch is clean), or we are returning 0 from `git status --porcelain | grep .` which triggers `&& $(FAIL)` and returns `1` (when the branch is dirty). I fee like a cleaner solution could be:
```
if git status --porcelain | grep . ; then $(FAIL); else $(OK) branch is clean; fi
```

How does that look to you? I could be convinced that calling `git status -s` is more obvious than the `grep .` we are doing.

haarchri: @hasheddan  tested your code snip - yes it is cleaner - added in the MR - thanks

final test looks as expected 

```
20:33:16 [ .. ] checking that branch is clean
20:33:16 [ OK ] branch is clean
```

```
20:36:44 [ .. ] checking that branch is clean
 M Testfile
20:36:44 [FAIL]
make: *** [check-diff] Error 1
```

hasheddan: Nice! Thanks @haarchri!

hasheddan (approved):

PR #169, opened 2021-09-01
Small utility to extract PR list to compose release notes

muvaf:

This is basically prints all merged PRs with their number and title. We can further improve it to classify them like new features, bug fixes etc.

I have:

Manually with `release-1.4` branch of Crossplane during 1.4 release.


scripts/release-notes.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import os
```
muvaf, line 4: It's been ages since I wrote Python code, so don't spare any comments. I have no strong opinion at all.

muvaf: One thing I tried hard to do is to use built-in modules so that we don't have to have a `requirements.txt` or deal with env-dependent things for such a small utility.

muvaf:
Updated the variable names to `BASE` and `RELEASE` because you can use tag, commit id or branch as long as it's something `git` can `checkout`.

PR #173, opened 2021-11-24
Add imagelight as alternative image building orchestrator

hasheddan:

Adds a new image building orchestration Makefile that is mutually
exclusive with image.mk. The purpose of imagelight.mk is to defer more
of the functionality to the individual image makefiles that exist in a
repo that consumes the build submodule. This use case has become more
desirable as we move towards tools such as buildx, but imagelight does
not specifically constrain the consumer to any one image building tool.

There are a few notable differences from image.mk, including but not
limited to:
- We do not save images as tarballs to an output directory, though
individual image makefiles in a repo could opt to do so on their own.
- We support publishing to multiple registries and repositories. An
arbitrary number of REGISTRY_ORGS can be defined in the top-level
Makefile in a repo, and each image defined in that repository will be
published to every specified location in REGISTRY_ORGS. REGISTRY_ORGS
take the format of registry-name/org-name (e.g. docker.io/crossplane).
- We skip building for platforms that are not valid, which, for now,
excludes all non-linux platforms. This simplifies the process of
specifying platforms in a repository because we do not force the
consumer to override platforms for which they want to build binaries but
do not want to or cannot publish images.
- We do not call common.mk targets when invoking the image-specific
Makefiles in a repository. This is an unnecessary level of indirection
that doesn't provide much additional value in any of the repos that
consume the build submodule. We instead call the targets in the image
Makefiles directly, requiring img.build, img.publish, and img.promote to
be defined.

Note that image.mk remain unmodified, which allows for any repositories
that currently consume it to continue doing so. I do not anticipate
removing image.mk any time soon as it provides different functionality
than imagelight.mk and some projects may opt never to switch.

I have:

I have tested this functionality fairly extensively with building / publishing / promoting images in `crossplane/crossplane`. I will also open a PR there to demonstrate this functionality, and will link it here. For reference, here is an example image Makefile that opts to use `docker buildx` for image operations.

```
# ====================================================================================
# Setup Project

include ../../../build/makelib/common.mk

# ====================================================================================
#  Options

include ../../../build/makelib/imagelight.mk

# ====================================================================================
# Targets

img.build:
	@$(INFO) docker build $(IMAGE)
	@$(MAKE) BUILD_ARGS="--load" img.build.shared
	@$(OK) docker build $(IMAGE)

img.publish:
	@$(INFO) docker publish $(IMAGE)
	@$(MAKE) BUILD_ARGS="--push" img.build.shared
	@$(OK) docker publish $(IMAGE)

img.build.shared:
	@cp Dockerfile $(IMAGE_TEMP_DIR) || $(FAIL)
	@cp -r $(OUTPUT_DIR)/bin/ $(IMAGE_TEMP_DIR)/bin || $(FAIL)
	@cp -a ../../../cluster/crds $(IMAGE_TEMP_DIR) || $(FAIL)
	@docker buildx build $(BUILD_ARGS) \
		--platform $(IMAGE_PLATFORMS) \
		-t $(IMAGE) \
		$(IMAGE_TEMP_DIR) || $(FAIL)

img.promote:
	@$(INFO) docker promote $(FROM_IMAGE) to $(TO_IMAGE)
	@docker buildx imagetools create -t $(TO_IMAGE) $(FROM_IMAGE)
	@$(OK) docker promote $(FROM_IMAGE) to $(TO_IMAGE)
```

### DIFF
--- a/makelib/imagelight.mk
+++ b/makelib/imagelight.mk
@@ -1,0 +1,214 @@
+# Copyright 2021 The Upbound Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ====================================================================================
+# Options
+
+ifeq ($(origin IMAGE_DIR),undefined)
+IMAGE_DIR := $(ROOT_DIR)/cluster/images
+endif
+
+ifeq ($(origin IMAGE_OUTPUT_DIR),undefined)
+IMAGE_OUTPUT_DIR := $(OUTPUT_DIR)/images/$(PLATFORM)
+endif
+
+ifeq ($(origin IMAGE_TEMP_DIR),undefined)
+IMAGE_TEMP_DIR := $(shell mktemp -d)
+endif
+
+# a registry that is scoped to the current build tree on this host. this enables
+# us to have isolation between concurrent builds on the same system, as in the case
+# of multiple working directories or on a CI system with multiple executors. All images
+# tagged with this build registry can safely be untagged/removed at the end of the build.
+ifeq ($(origin BUILD_REGISTRY), undefined)
+BUILD_REGISTRY := build-$(shell echo $(HOSTNAME)-$(ROOT_DIR) | shasum -a 256 | cut -c1-8)
+endif
+
+# In order to reduce built time especially on jenkins, we maintain a cache
+# of already built images. This cache contains images that can be used to help speed
+# future docker build commands using docker's content addressable schemes.
+# All cached images go in in a 'cache/' local registry and we follow an MRU caching
+# policy -- keeping images that have been referenced around and evicting images
+# that have to been referenced in a while (and according to a policy). Note we can
+# not rely on the image's .CreatedAt date since docker only updates then when the
+# image is created and not referenced. Instead we keep a date in the Tag.
+CACHE_REGISTRY := cache
+
+# prune images that are at least this many hours old
+PRUNE_HOURS ?= 48
+
+# prune keeps at least this many images regardless of how old they are
+PRUNE_KEEP ?= 24
+
+# don't actually prune just show what prune would do.
+PRUNE_DRYRUN ?= 0
+
+# the cached image format
+CACHE_DATE_FORMAT := "%Y-%m-%d.%H%M%S"
+CACHE_PRUNE_DATE := $(shell export TZ="UTC+$(PRUNE_HOURS)"; date +"$(CACHE_DATE_FORMAT)")
+CACHE_TAG := $(shell date -u +"$(CACHE_DATE_FORMAT)")
+
+REGISTRY_ORGS ?= docker.io
+IMAGE_ARCHS := $(subst linux_,,$(filter linux_%,$(PLATFORMS)))
+IMAGE_PLATFORMS := $(subst _,/,$(subst $(SPACE),$(COMMA),$(filter linux_%,$(PLATFORMS))))
+IMAGE_PLATFORMS_LIST := $(subst _,/,$(filter linux_%,$(PLATFORMS)))
+IMAGE_PLATFORM := $(subst _,/,$(PLATFORM))
+
+# if set to 1 docker image caching will not be used.
+CACHEBUST ?= 0
+ifeq ($(CACHEBUST),1)
+BUILD_ARGS += --no-cache
+endif
+
+ifeq ($(HOSTOS),Linux)
+SELF_CID := $(shell cat /proc/self/cgroup | grep docker | grep -o -E '[0-9a-f]{64}' | head -n 1)
+endif
+
+# =====================================================================================
+# Image Targets
+
+do.img.clean:
+	@for i in $(CLEAN_IMAGES); do \
+		if [ -n "$$(docker images -q $$i)" ]; then \
+			for c in $$(docker ps -a -q --no-trunc --filter=ancestor=$$i); do \
+				if [ "$$c" != "$(SELF_CID)" ]; then \
+					echo stopping and removing container $${c} referencing image $$i; \
+					docker stop $${c}; \
+					docker rm $${c}; \
+				fi; \
+			done; \
+			echo cleaning image $$i; \
+			docker rmi $$i > /dev/null 2>&1 || true; \
+		fi; \
+	done
+
+# this will clean everything for this build
+img.clean:
+	@$(INFO) cleaning images for $(BUILD_REGISTRY)
+	@$(MAKE) do.img.clean CLEAN_IMAGES="$(shell docker images | grep -E '^$(BUILD_REGISTRY)/' | awk '{print $$1":"$$2}')"
+	@$(OK) cleaning images for $(BUILD_REGISTRY)
+
+img.done:
+	@rm -fr $(IMAGE_TEMP_DIR)
+
+img.cache:
+	@for i in $(CACHE_IMAGES); do \
+		IMGID=$$(docker images -q $$i); \
+		if [ -n "$$IMGID" ]; then \
+			echo === caching image $$i; \
+			CACHE_IMAGE=$(CACHE_REGISTRY)/$${i#*/}; \
+			docker tag $$i $${CACHE_IMAGE}:$(CACHE_TAG); \
+			for r in $$(docker images --format "{{.ID}}#{{.Repository}}:{{.Tag}}" | grep $$IMGID | grep $(CACHE_REGISTRY)/ | grep -v $${CACHE_IMAGE}:$(CACHE_TAG)); do \
+				docker rmi $${r#*#} > /dev/null 2>&1 || true; \
+			done; \
+		fi; \
+	done
+
+# prune removes old cached images
+img.prune:
+	@$(INFO) pruning images older than $(PRUNE_HOURS) keeping a minimum of $(PRUNE_KEEP) images
+	@EXPIRED=$$(docker images --format "{{.Tag}}#{{.Repository}}:{{.Tag}}" \
+		| grep -E '$(CACHE_REGISTRY)/' \
+		| sort -r \
+		| awk -v i=0 -v cd="$(CACHE_PRUNE_DATE)" -F  "#" '{if ($$1 <= cd && i >= $(PRUNE_KEEP)) print $$2; i++ }') &&\
+	for i in $$EXPIRED; do \
+		echo removing expired cache image $$i; \
+		[ $(PRUNE_DRYRUN) = 1 ] || docker rmi $$i > /dev/null 2>&1 || true; \
+	done
+	@for i in $$(docker images -q -f dangling=true); do \
+		echo removing dangling image $$i; \
+		docker rmi $$i > /dev/null 2>&1 || true; \
+	done
+	@$(OK) pruning
+
+debug.nuke:
+	@for c in $$(docker ps -a -q --no-trunc); do \
+		if [ "$$c" != "$(SELF_CID)" ]; then \
+			echo stopping and removing container $${c}; \
+			docker stop $${c}; \
+			docker rm $${c}; \
+		fi; \
+	done
+	@for i in $$(docker images -q); do \
+		echo removing image $$i; \
+		docker rmi -f $$i > /dev/null 2>&1; \
+	done
+
+# 1: registry 2: image
+define repo.targets
+img.release.publish.$(1).$(2):
+	@$(MAKE) -C $(IMAGE_DIR)/$(2) IMAGE_PLATFORMS=$(IMAGE_PLATFORMS) IMAGE=$(1)/$(2):$(VERSION) img.publish
+img.release.publish: img.release.publish.$(1).$(2)
+
+img.release.promote.$(1).$(2):
+	@$(MAKE) -C $(IMAGE_DIR)/$(2) TO_IMAGE=$(1)/$(2):$(CHANNEL) FROM_IMAGE=$(1)/$(2):$(VERSION) img.promote
+	@[ "$(CHANNEL)" = "master" ] || @$(MAKE) -C $(IMAGE_DIR)/$(2) TO_IMAGE=$(1)/$(2):$(VERSION)-$(CHANNEL) FROM_IMAGE=$(1)/$(2):$(VERSION) img.promote
+img.release.promote: img.release.promote.$(1).$(2)
+
+img.release.clean.$(1).$(2):
+	@[ -z "$$$$(docker images -q $(1)/$(2):$(VERSION))" ] || docker rmi $(1)/$(2):$(VERSION)
+	@[ -z "$$$$(docker images -q $(1)/$(2):$(VERSION)-$(CHANNEL))" ] || docker rmi $(1)/$(2):$(VERSION)-$(CHANNEL)
+	@[ -z "$$$$(docker images -q $(1)/$(2):$(CHANNEL))" ] || docker rmi $(1)/$(2):$(CHANNEL)
+img.release.clean: img.release.clean.$(1).$(2)
+endef
+$(foreach r,$(REGISTRY_ORGS), $(foreach i,$(IMAGES),$(eval $(call repo.targets,$(r),$(i)))))
+
+# ====================================================================================
+# Common Targets
+
+do.build.image.%:
+	@$(MAKE) -C $(IMAGE_DIR)/$* IMAGE_PLATFORMS=$(IMAGE_PLATFORM) IMAGE=$(BUILD_REGISTRY)/$*-$(ARCH) img.build
+do.build.images: $(foreach i,$(IMAGES), do.build.image.$(i))
+do.skip.images:
+	@$(OK) Skipping image build for unsupported platform $(IMAGE_PLATFORM)
+
+ifneq ($(filter $(IMAGE_PLATFORM),$(IMAGE_PLATFORMS_LIST)),)
+build.artifacts.platform: do.build.images
+else
+build.artifacts.platform: do.skip.images
+endif
+build.done: img.cache img.done
+clean: img.clean img.release.clean
+
+# only publish images for main / master and release branches
+# TODO(hasheddan): remove master and support overriding
+ifneq ($(filter main master release-%,$(BRANCH_NAME)),)
+publish.artifacts: $(foreach r,$(REGISTRY_ORGS), $(foreach i,$(IMAGES),img.release.publish.$(r).$(i)))
+endif
+
+promote.artifacts: $(foreach r,$(REGISTRY_ORGS), $(foreach i,$(IMAGES),img.release.promote.$(r).$(i)))
+
+# ====================================================================================
+# Special Targets
+
+prune: img.prune
+
+define IMAGE_HELPTEXT
+Image Targets:
+    prune        Prune orphaned and cached images.
+
+Image Options:
+    PRUNE_HOURS  The number of hours from when an image is last used for it to be
+                 considered a target for pruning. Default is 48 hrs.
+    PRUNE_KEEP   The minimum number of cached images to keep. Default is 24 images.
+
+endef
+export IMAGE_HELPTEXT
+
+img.help:
+	@echo "$$IMAGE_HELPTEXT"
+
+help-special: img.help
+
+.PHONY: prune img.help
--- a/makelib/imagelight.mk
+++ b/makelib/imagelight.mk
@@ -153,7 +153,7 @@ img.release.publish: img.release.publish.$(1).$(2)
 
 img.release.promote.$(1).$(2):
 	@$(MAKE) -C $(IMAGE_DIR)/$(2) TO_IMAGE=$(1)/$(2):$(CHANNEL) FROM_IMAGE=$(1)/$(2):$(VERSION) img.promote
-	@[ "$(CHANNEL)" = "master" ] || @$(MAKE) -C $(IMAGE_DIR)/$(2) TO_IMAGE=$(1)/$(2):$(VERSION)-$(CHANNEL) FROM_IMAGE=$(1)/$(2):$(VERSION) img.promote
+	@[ "$(CHANNEL)" = "master" ] || $(MAKE) -C $(IMAGE_DIR)/$(2) TO_IMAGE=$(1)/$(2):$(VERSION)-$(CHANNEL) FROM_IMAGE=$(1)/$(2):$(VERSION) img.promote
 img.release.promote: img.release.promote.$(1).$(2)
 
 img.release.clean.$(1).$(2):
--- a/makelib/imagelight.mk
+++ b/makelib/imagelight.mk
@@ -35,30 +35,6 @@ ifeq ($(origin BUILD_REGISTRY), undefined)
 BUILD_REGISTRY := build-$(shell echo $(HOSTNAME)-$(ROOT_DIR) | shasum -a 256 | cut -c1-8)
 endif
 
-# In order to reduce built time especially on jenkins, we maintain a cache
-# of already built images. This cache contains images that can be used to help speed
-# future docker build commands using docker's content addressable schemes.
-# All cached images go in in a 'cache/' local registry and we follow an MRU caching
-# policy -- keeping images that have been referenced around and evicting images
-# that have to been referenced in a while (and according to a policy). Note we can
-# not rely on the image's .CreatedAt date since docker only updates then when the
-# image is created and not referenced. Instead we keep a date in the Tag.
-CACHE_REGISTRY := cache
-
-# prune images that are at least this many hours old
-PRUNE_HOURS ?= 48
-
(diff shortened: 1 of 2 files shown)


hasheddan (commented):
Aforementioned `crossplane/crossplane` PR: https://github.com/crossplane/crossplane/pull/2718

makelib/imagelight.mk
```diff
@@ -0,0 +1,214 @@
...
+# future docker build commands using docker's content addressable schemes.
+# All cached images go in in a 'cache/' local registry and we follow an MRU caching
+# policy -- keeping images that have been referenced around and evicting images
+# that have to been referenced in a while (and according to a policy). Note we can
+# not rely on the image's .CreatedAt date since docker only updates then when the
+# image is created and not referenced. Instead we keep a date in the Tag.
+CACHE_REGISTRY := cache
+
+# prune images that are at least this many hours old
+PRUNE_HOURS ?= 48
```
tnthornton, line 49: Couple of questions regarding the cache:
1. How do we plan to utilize the images that we're retagging to keep in the 'cache'?
I don't currently see anywhere where we're utilizing `cache-from` in the buildx commands? I see the `--load` arg being used [here](https://github.com/crossplane/crossplane/blob/e335e7c077d0269e6e1462eab44892c566e51be2/cluster/images/crossplane/Makefile#L17), however I think that's only effective for [single-platform builds](https://docs.docker.com/engine/reference/commandline/buildx_build/#load).

2. With `PRUNE_HOURS` set to 48, are there any concerns with the cache being only effective on subsequent branch builds, instead of being available for first build (for example when running on main/master)?
It seems like we'd have multiple items at play here:
- The build system (GH Actions, Circle, Travis, etc) and how it enables Docker build caching. As an example, Circle can be pretty hit or miss in what it has available in DLC. [ref](https://circleci.com/docs/2.0/docker-layer-caching/#how-dlc-works).
- How frequently builds are running. If a build only runs once every couple of days, it seems like we'll be starting from a "fresh" cache each time.

makelib/imagelight.mk
```diff
@@ -0,0 +1,214 @@
...
+img.clean:
+	@$(INFO) cleaning images for $(BUILD_REGISTRY)
+	@$(MAKE) do.img.clean CLEAN_IMAGES="$(shell docker images | grep -E '^$(BUILD_REGISTRY)/' | awk '{print $$1":"$$2}')"
+	@$(OK) cleaning images for $(BUILD_REGISTRY)
+
+img.done:
+	@rm -fr $(IMAGE_TEMP_DIR)
+
+img.cache:
+	@for i in $(CACHE_IMAGES); do \
```
tnthornton, line 106: Does this mean that we'll need to define the list of `CACHE_IMAGES` in the top level Makefile for a project?

makelib/imagelight.mk
```diff
@@ -0,0 +1,214 @@
...
+img.release.clean: img.release.clean.$(1).$(2)
+endef
+$(foreach r,$(REGISTRY_ORGS), $(foreach i,$(IMAGES),$(eval $(call repo.targets,$(r),$(i)))))
+
+# ====================================================================================
+# Common Targets
+
+do.build.image.%:
+	@$(MAKE) -C $(IMAGE_DIR)/$* IMAGE_PLATFORMS=$(IMAGE_PLATFORM) IMAGE=$(BUILD_REGISTRY)/$*-$(ARCH) img.build
+do.build.images: $(foreach i,$(IMAGES), do.build.image.$(i))
```
tnthornton, line 172: So this piece may answer part of my "how are we utilizing the cache" question. It looks like we're narrowing down to one platform at a time to build the images. Given that, I am curious if `buildx` will experience the same problem we saw with `docker build` and the mutli-arch static container when running parallel image builds for each target platform.

makelib/imagelight.mk
hasheddan: @tnthornton these are all valid questions, to which my answer if mostly: We don't use the cache functionality here anywhere that I am aware of. The only reason it is included here is that it is 100% copied from the existing [image.mk](https://github.com/upbound/build/blob/master/makelib/image.mk), and I tried to keep everything that wouldn't actually prohibit us from using something like `buildx` (i.e. if the functionality is already present and it doesn't disrupt what we are trying to do then keep it available for consumers). That being said, I also considered stripping out everything that we don't need to make this as "light" as possible. Do you think you would advocate for that approach?

makelib/imagelight.mk
hasheddan: I believe if someone was consuming this functionality then that is true. As mentioned above, I am not aware of any of our projects that are doing so.

makelib/imagelight.mk
hasheddan: @tnthornton this does call `img.build` individually for each platform for each image 👍🏻 I am fairly certain that we will not experience this race condition with `buildx` as the underlying builders are intended to be able to [handle parallel requests](https://github.com/docker/buildx/issues/344#issuecomment-670152753). This is essentially [how `docker buildx bake` works](https://github.com/docker/buildx/blob/master/docs/reference/buildx_bake.md). This is aligned with the tests I have ran with parallelism as well.

One other thing to note here is that we do run individual builds to load into the daemon for local dev and CI, but when we push we use `--platforms` and `--push` to build a single mutli-arch image (see example Makefile in PR body).

hasheddan (commented):
Thanks for your review @tnthornton! Let me know if you think that we should clean out some of the `cache` portions of this -- I think that is a valid outlook, and we could always add it back if needed in the future :)

makelib/imagelight.mk
tnthornton: @hasheddan sorry for the slow response.

>The only reason it is included here is that it is 100% copied from the existing image.mk, and I tried to keep everything that wouldn't actually prohibit us from using something like buildx (i.e. if the functionality is already present and it doesn't disrupt what we are trying to do then keep it available for consumers).

That makes sense to me and I'd probably take the same approach.

>That being said, I also considered stripping out everything that we don't need to make this as "light" as possible. Do you think you would advocate for that approach?

I think currently, yes(?). Especially if this was copied from the existing `image.mk`. Given that, if we find ourselves wanting to use cache pruning, etc we can bring it back or reference that impl.

FWIW, I do find myself going back and forth about this. So if one of the other reviewers find that they feel more strongly about just keeping it in here for now, I think I can be convinced to get behind that.

makelib/imagelight.mk
```diff
@@ -0,0 +1,214 @@
...
+img.prune:
+	@$(INFO) pruning images older than $(PRUNE_HOURS) keeping a minimum of $(PRUNE_KEEP) images
+	@EXPIRED=$$(docker images --format "{{.Tag}}#{{.Repository}}:{{.Tag}}" \
+		| grep -E '$(CACHE_REGISTRY)/' \
+		| sort -r \
+		| awk -v i=0 -v cd="$(CACHE_PRUNE_DATE)" -F  "#" '{if ($$1 <= cd && i >= $(PRUNE_KEEP)) print $$2; i++ }') &&\
+	for i in $$EXPIRED; do \
+		echo removing expired cache image $$i; \
+		[ $(PRUNE_DRYRUN) = 1 ] || docker rmi $$i > /dev/null 2>&1 || true; \
+	done
```
hasheddan, line 128: I noticed that the cross image uses this cache functionality, so some of this pruning logic would skip that now -- however, part of the point of this change is that we are no longer building the cross image, so I am not too concerned about removing all of the existing cache stuff

makelib/imagelight.mk
hasheddan: I think I tend to agree -- part of the motivation for this change is our current process is "doing too much", so if we are not using something might as well remove it and not allow the potential for someone depending on it, then trying to remove it

negz: > Do you think you would advocate for that approach?

My intuition is also yes here. If this is going to be an alternative to the existing process it makes sense to me to take the opportunity to split out anything superfluous and 'start from scratch' rather than effectively forking what we already have an modifying it.

negz:
> Note that image.mk remain unmodified, which allows for any repositories that currently consume it to continue doing so. I do not anticipate removing image.mk any time soon as it provides different functionality than imagelight.mk and some projects may opt never to switch.

@hasheddan Can you elaborate on this? My preference would definitely be to pick the variant we like better and at least slowly align on that. Is there a reason a project would prefer one image library or the other? If we were setting up a new project is there a reason we might recommend `image.mk` over the new variant introduced here?

makelib/imagelight.mk
negz: Side note - if we do keep all the stuff that was copied from `image.mk` a diff between `image.mk` and `imagelight.mk` would help reviewers focus on what's new here (tough to do though since there's no `git cp` - maybe just a `git diff` in a comment? 😬)

hasheddan:
> My preference would definitely be to pick the variant we like better and at least slowly align on that. Is there a reason a project would prefer one image library or the other? If we were setting up a new project is there a reason we might recommend image.mk over the new variant introduced here?

@negz I would definitely recommend aligning on `imagelight.mk`, but I am trying to make allowances for projects that already use `image.mk` (which could be an arbitrary number of projects since this is open source) because this is a fairly significant breaking change. That being said, if we do have strong signal that no one is utilizing `image.mk` anymore then I think it definitely makes sense to deprecate and then remove 👍🏻

hasheddan (commented):
@negz @tnthornton  `imagelight.mk` -> `image.mk` diff:

<details>
  <summary>Click to expand!</summary>




```
diff --git a/makelib/image.mk b/makelib/image.mk
index bcb1c18..b88ddbb 100644
--- a/makelib/image.mk
+++ b/makelib/image.mk
@@ -1,4 +1,4 @@
-# Copyright 2016 The Upbound Authors. All rights reserved.
+# Copyright 2021 The Upbound Authors. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -27,19 +27,6 @@ ifeq ($(origin IMAGE_TEMP_DIR),undefined)
 IMAGE_TEMP_DIR := $(shell mktemp -d)
 endif
 
-# set the OS base image to alpine if in not defined. set your own image for each
-# supported platform.
-ifeq ($(origin OSBASEIMAGE),undefined)
-OSBASE ?= alpine:3.13
-ifeq ($(ARCH),$(filter $(ARCH),amd64 ppc64le))
-OSBASEIMAGE = $(OSBASE)
-else ifeq ($(ARCH),arm64)
-OSBASEIMAGE = arm64v8/$(OSBASE)
-else
-$(error unsupported architecture $(ARCH))
-endif
-endif
-
 # a registry that is scoped to the current build tree on this host. this enables
 # us to have isolation between concurrent builds on the same system, as in the case
 # of multiple working directories or on a CI system with multiple executors. All images
@@ -48,36 +35,11 @@ ifeq ($(origin BUILD_REGISTRY), undefined)
 BUILD_REGISTRY := build-$(shell echo $(HOSTNAME)-$(ROOT_DIR) | shasum -a 256 | cut -c1-8)
 endif
 
-MANIFEST_TOOL_VERSION=v0.7.0
-MANIFEST_TOOL := $(TOOLS_HOST_DIR)/manifest-tool-$(MANIFEST_TOOL_VERSION)
-
-# In order to reduce built time especially on jenkins, we maintain a cache
-# of already built images. This cache contains images that can be used to help speed
-# future docker build commands using docker's content addressable schemes.
-# All cached images go in in a 'cache/' local registry and we follow an MRU caching
-# policy -- keeping images that have been referenced around and evicting images
-# that have to been referenced in a while (and according to a policy). Note we can
-# not rely on the image's .CreatedAt date since docker only updates then when the
-# image is created and not referenced. Instead we keep a date in the Tag.
-CACHE_REGISTRY := cache
-
-# prune images that are at least this many hours old
-PRUNE_HOURS ?= 48
-
-# prune keeps at least this many images regardless of how old they are
-PRUNE_KEEP ?= 24
-
-# don't actually prune just show what prune would do.
-PRUNE_DRYRUN ?= 0
-
-# the cached image format
-CACHE_DATE_FORMAT := "%Y-%m-%d.%H%M%S"
-CACHE_PRUNE_DATE := $(shell export TZ="UTC+$(PRUNE_HOURS)"; date +"$(CACHE_DATE_FORMAT)")
-CACHE_TAG := $(shell date -u +"$(CACHE_DATE_FORMAT)")
-
-REGISTRIES ?= $(DOCKER_REGISTRY)
+REGISTRY_ORGS ?= docker.io
 IMAGE_ARCHS := $(subst linux_,,$(filter linux_%,$(PLATFORMS)))
 IMAGE_PLATFORMS := $(subst _,/,$(subst $(SPACE),$(COMMA),$(filter linux_%,$(PLATFORMS))))
+IMAGE_PLATFORMS_LIST := $(subst _,/,$(filter linux_%,$(PLATFORMS)))
+IMAGE_PLATFORM := $(subst _,/,$(PLATFORM))
 
 # if set to 1 docker image caching will not be used.
 CACHEBUST ?= 0
@@ -85,22 +47,6 @@ ifeq ($(CACHEBUST),1)
 BUILD_ARGS += --no-cache
 endif
 
-# if V=0 avoid showing verbose output from docker build
-ifeq ($(V),0)
-BUILD_ARGS ?= -q
-endif
-
-# if PULL=1 we will always check if there is a newer base image
-PULL ?= 1
-ifeq ($(PULL),1)
-BUILD_BASE_ARGS += --pull
-endif
-BUILD_BASE_ARGS += $(BUILD_ARGS)
-export PULL
-
-# the version of tini to use
-TINI_VERSION ?= v0.16.1
-
 ifeq ($(HOSTOS),Linux)
 SELF_CID := $(shell cat /proc/self/cgroup | grep docker | grep -o -E '[0-9a-f]{64}' | head -n 1)
 endif
@@ -132,157 +78,46 @@ img.clean:
 img.done:
 	@rm -fr $(IMAGE_TEMP_DIR)
 
-img.cache:
-	@for i in $(CACHE_IMAGES); do \
-		IMGID=$$(docker images -q $$i); \
-		if [ -n "$$IMGID" ]; then \
-			echo === caching image $$i; \
-			CACHE_IMAGE=$(CACHE_REGISTRY)/$${i#*/}; \
-			docker tag $$i $${CACHE_IMAGE}:$(CACHE_TAG); \
-			for r in $$(docker images --format "{{.ID}}#{{.Repository}}:{{.Tag}}" | grep $$IMGID | grep $(CACHE_REGISTRY)/ | grep -v $${CACHE_IMAGE}:$(CACHE_TAG)); do \
-				docker rmi $${r#*#} > /dev/null 2>&1 || true; \
-			done; \
-		fi; \
-	done
-
-# prune removes old cached images
-img.prune:
-	@$(INFO) pruning images older than $(PRUNE_HOURS) keeping a minimum of $(PRUNE_KEEP) images
-	@EXPIRED=$$(docker images --format "{{.Tag}}#{{.Repository}}:{{.Tag}}" \
-		| grep -E '$(CACHE_REGISTRY)/' \
-		| sort -r \
-		| awk -v i=0 -v cd="$(CACHE_PRUNE_DATE)" -F  "#" '{if ($$1 <= cd && i >= $(PRUNE_KEEP)) print $$2; i++ }') &&\
-	for i in $$EXPIRED; do \
-		echo removing expired cache image $$i; \
-		[ $(PRUNE_DRYRUN) = 1 ] || docker rmi $$i > /dev/null 2>&1 || true; \
-	done
-	@for i in $$(docker images -q -f dangling=true); do \
-		echo removing dangling image $$i; \
-		docker rmi $$i > /dev/null 2>&1 || true; \
-	done
-	@$(OK) pruning
-
-debug.nuke:
-	@for c in $$(docker ps -a -q --no-trunc); do \
-		if [ "$$c" != "$(SELF_CID)" ]; then \
-			echo stopping and removing container $${c}; \
-			docker stop $${c}; \
-			docker rm $${c}; \
-		fi; \
-	done
-	@for i in $$(docker images -q); do \
-		echo removing image $$i; \
-		docker rmi -f $$i > /dev/null 2>&1; \
-	done
-
-# 1: registry 2: image, 3: arch
+# 1: registry 2: image
 define repo.targets
-img.release.build.$(1).$(2).$(3):
-	@$(INFO) docker build $(1)/$(2)-$(3):$(VERSION)
-	@docker tag $(BUILD_REGISTRY)/$(2)-$(3) $(1)/$(2)-$(3):$(VERSION) || $(FAIL)
-	@# Save image as _output/images/linux_<arch>/<image>.tar.gz (no builds for darwin or windows)
-	@mkdir -p $(OUTPUT_DIR)/images/linux_$(3) || $(FAIL)
-	@docker save $(BUILD_REGISTRY)/$(2)-$(3) | gzip -c > $(OUTPUT_DIR)/images/linux_$(3)/$(2).tar.gz || $(FAIL)
-	@$(OK) docker build $(1)/$(2)-$(3):$(VERSION)
-img.release.build: img.release.build.$(1).$(2).$(3)
-
-img.release.publish.$(1).$(2).$(3):
-	@$(INFO) docker push $(1)/$(2)-$(3):$(VERSION)
-	@docker push $(1)/$(2)-$(3):$(VERSION) || $(FAIL)
-	@$(OK) docker push $(1)/$(2)-$(3):$(VERSION)
-img.release.publish: img.release.publish.$(1).$(2).$(3)
-
-img.release.promote.$(1).$(2).$(3):
-	@$(INFO) docker promote $(1)/$(2)-$(3):$(VERSION) to $(1)/$(2)-$(3):$(CHANNEL)
-	@docker pull $(1)/$(2)-$(3):$(VERSION) || $(FAIL)
-	@[ "$(CHANNEL)" = "master" ] || docker tag $(1)/$(2)-$(3):$(VERSION) $(1)/$(2)-$(3):$(VERSION)-$(CHANNEL) || $(FAIL)
-	@docker tag $(1)/$(2)-$(3):$(VERSION) $(1)/$(2)-$(3):$(CHANNEL) || $(FAIL)
-	@[ "$(CHANNEL)" = "master" ] || docker push $(1)/$(2)-$(3):$(VERSION)-$(CHANNEL)
-	@docker push $(1)/$(2)-$(3):$(CHANNEL) || $(FAIL)
-	@$(OK) docker promote $(1)/$(2)-$(3):$(VERSION) to $(1)/$(2)-$(3):$(CHANNEL) || $(FAIL)
-img.release.promote: img.release.promote.$(1).$(2).$(3)
-
-img.release.clean.$(1).$(2).$(3):
-	@[ -z "$$$$(docker images -q $(1)/$(2)-$(3):$(VERSION))" ] || docker rmi $(1)/$(2)-$(3):$(VERSION)
-	@[ -z "$$$$(docker images -q $(1)/$(2)-$(3):$(VERSION)-$(CHANNEL))" ] || docker rmi $(1)/$(2)-$(3):$(VERSION)-$(CHANNEL)
-	@[ -z "$$$$(docker images -q $(1)/$(2)-$(3):$(CHANNEL))" ] || docker rmi $(1)/$(2)-$(3):$(CHANNEL)
-img.release.clean: img.release.clean.$(1).$(2).$(3)
+img.release.publish.$(1).$(2):
+	@$(MAKE) -C $(IMAGE_DIR)/$(2) IMAGE_PLATFORMS=$(IMAGE_PLATFORMS) IMAGE=$(1)/$(2):$(VERSION) img.publish
+img.release.publish: img.release.publish.$(1).$(2)
+
+img.release.promote.$(1).$(2):
+	@$(MAKE) -C $(IMAGE_DIR)/$(2) TO_IMAGE=$(1)/$(2):$(CHANNEL) FROM_IMAGE=$(1)/$(2):$(VERSION) img.promote
+	@[ "$(CHANNEL)" = "master" ] || $(MAKE) -C $(IMAGE_DIR)/$(2) TO_IMAGE=$(1)/$(2):$(VERSION)-$(CHANNEL) FROM_IMAGE=$(1)/$(2):$(VERSION) img.promote
+img.release.promote: img.release.promote.$(1).$(2)
+
+img.release.clean.$(1).$(2):
+	@[ -z "$$$$(docker images -q $(1)/$(2):$(VERSION))" ] || docker rmi $(1)/$(2):$(VERSION)
+	@[ -z "$$$$(docker images -q $(1)/$(2):$(VERSION)-$(CHANNEL))" ] || docker rmi $(1)/$(2):$(VERSION)-$(CHANNEL)
+	@[ -z "$$$$(docker images -q $(1)/$(2):$(CHANNEL))" ] || docker rmi $(1)/$(2):$(CHANNEL)
+img.release.clean: img.release.clean.$(1).$(2)
 endef
-$(foreach r,$(REGISTRIES), $(foreach i,$(IMAGES), $(foreach a,$(IMAGE_ARCHS),$(eval $(call repo.targets,$(r),$(i),$(a))))))
-
-img.release.manifest.publish.%: img.release.publish $(MANIFEST_TOOL)
-	@$(MANIFEST_TOOL) push from-args --platforms $(IMAGE_PLATFORMS) --template $(DOCKER_REGISTRY)/$*-ARCH:$(VERSION) --target $(DOCKER_REGISTRY)/$*:$(VERSION) || $(FAIL)
-
-img.release.manifest.promote.%: img.release.promote $(MANIFEST_TOOL)
-	@[ "$(CHANNEL)" = "master" ] || $(MANIFEST_TOOL) push from-args --platforms $(IMAGE_PLATFORMS) --template $(DOCKER_REGISTRY)/$*-ARCH:$(VERSION) --target $(DOCKER_REGISTRY)/$*:$(VERSION)-$(CHANNEL) || $(FAIL)
-	@$(MANIFEST_TOOL) push from-args --platforms $(IMAGE_PLATFORMS) --template $(DOCKER_REGISTRY)/$*-ARCH:$(VERSION) --target $(DOCKER_REGISTRY)/$*:$(CHANNEL) || $(FAIL)
+$(foreach r,$(REGISTRY_ORGS), $(foreach i,$(IMAGES),$(eval $(call repo.targets,$(r),$(i)))))
 
 # ====================================================================================
 # Common Targets
 
-# if IMAGES is defined then invoke and build each image identified
-ifneq ($(IMAGES),)
+do.build.image.%:
+	@$(MAKE) -C $(IMAGE_DIR)/$* IMAGE_PLATFORMS=$(IMAGE_PLATFORM) IMAGE=$(BUILD_REGISTRY)/$*-$(ARCH) img.build
+do.build.images: $(foreach i,$(IMAGES), do.build.image.$(i))
+do.skip.images:
+	@$(OK) Skipping image build for unsupported platform $(IMAGE_PLATFORM)
 
-ifeq ($(DOCKER_REGISTRY),)
-$(error the variable DOCKER_REGISTRY must be set prior to including image.mk)
-endif
-
-do.build.image.%: ; @$(MAKE) -C $(IMAGE_DIR)/$* PLATFORM=$(PLATFORM)
-do.build.images: $(foreach i,$(IMAGES), do.build.image.$(i)) ;
+ifneq ($(filter $(IMAGE_PLATFORM),$(IMAGE_PLATFORMS_LIST)),)
 build.artifacts.platform: do.build.images
+else
+build.artifacts.platform: do.skip.images
+endif
 build.done: img.cache img.done
 clean: img.clean img.release.clean
 
-publish.init: img.release.build
-
 # only publish images for main / master and release branches
 # TODO(hasheddan): remove master and support overriding
 ifneq ($(filter main master release-%,$(BRANCH_NAME)),)
-publish.artifacts: $(addprefix img.release.manifest.publish.,$(IMAGES))
-endif
-
-promote.artifacts: $(addprefix img.release.manifest.promote.,$(IMAGES))
-
-else
-
-# otherwise we assume this .mk file is being included to build a single image
-
-build.artifacts.platform: img.build
-build.done: img.cache img.done
-clean: img.clean
-
+publish.artifacts: $(foreach r,$(REGISTRY_ORGS), $(foreach i,$(IMAGES),img.release.publish.$(r).$(i)))
 endif
 
-# ====================================================================================
-# Special Targets
-
-prune: img.prune
-
-define IMAGE_HELPTEXT
-Image Targets:
-    prune        Prune orphaned and cached images.
-
-Image Options:
-    PRUNE_HOURS  The number of hours from when an image is last used for it to be
-                 considered a target for pruning. Default is 48 hrs.
-    PRUNE_KEEP   The minimum number of cached images to keep. Default is 24 images.
-
-endef
-export IMAGE_HELPTEXT
-
-img.help:
-	@echo "$$IMAGE_HELPTEXT"
-
-help-special: img.help
-
-.PHONY: prune img.help
-
-# ====================================================================================
-# tools
-
-$(MANIFEST_TOOL):
-	@$(INFO) installing manifest-tool $(MANIFEST_TOOL_VERSION)
-	@mkdir -p $(TOOLS_HOST_DIR) || $(FAIL)
-	@curl -fsSL https://github.com/estesp/manifest-tool/releases/download/$(MANIFEST_TOOL_VERSION)/manifest-tool-$(HOSTOS)-$(SAFEHOSTARCH) > $@ || $(FAIL)
-	@chmod +x $@ || $(FAIL)
-	@$(OK) installing manifest-tool $(MANIFEST_TOOL_VERSION)
+promote.artifacts: $(foreach r,$(REGISTRY_ORGS), $(foreach i,$(IMAGES),img.release.promote.$(r).$(i)))

```

</details>

makelib/imagelight.mk
hasheddan: I have removed the caching / pruning / debug components, which I don't see us using anywhere of consequence. Also added a diff in the review at the bottom of the thread :)

tnthornton (approved):

negz:
> That being said, if we do have strong signal that no one is utilizing image.mk anymore then I think it definitely makes sense to deprecate and then remove

I don't think we need to wait for folks to not be using it before we mark it deprecated. Given we prefer folks to use this newer method, let's mark it deprecated now. We don't need to necessarily have a timeline or otherwise put any real effort into porting existing consumers to make it go away, but we should make it as easy and obvious as possible for new projects to make the "right choice" (which seems to be this new implementation).

PR #177, opened 2022-01-06
Update go to 1.17

muvaf:

Updates Go to 1.17.5

I have:

N/A

### DIFF
--- a/cross/Dockerfile
+++ b/cross/Dockerfile
@@ -45,8 +45,8 @@ RUN curl -fsSL "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-2.1.21.zip"
 
 # ------------------------------------------------------------------------------------------------
 # Go support
-RUN GO_VERSION=1.16.1 && \
-    GO_HASH=3edc22f8332231c3ba8be246f184b736b8d28f06ce24f08168d8ecf052549769 && \
+RUN GO_VERSION=1.17.5 && \
+    GO_HASH=bd78114b0d441b029c8fe0341f4910370925a4d270a6a590668840675b0c653e && \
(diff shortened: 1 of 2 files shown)


cross/Dockerfile
```diff
@@ -45,7 +45,7 @@ RUN curl -fsSL "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-2.1.21.zip"
 
 # ------------------------------------------------------------------------------------------------
 # Go support
-RUN GO_VERSION=1.16.1 && \
+RUN GO_VERSION=1.17.5 && \
     GO_HASH=3edc22f8332231c3ba8be246f184b736b8d28f06ce24f08168d8ecf052549769 && \
```
hasheddan, line 49: @muvaf I believe this hash also needs to be updated

cross/Dockerfile
```diff
@@ -45,7 +45,7 @@ RUN curl -fsSL "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-2.1.21.zip"
 
 # ------------------------------------------------------------------------------------------------
 # Go support
-RUN GO_VERSION=1.16.1 && \
+RUN GO_VERSION=1.17.5 && \
```
hasheddan, line 48: Just curious, did you have a case where you were running into issues with Go version? I believe we have moved away from using cross container everywhere in Crossplane ecosystem

cross/Dockerfile
muvaf: Thanks for catching!

hasheddan (approved):

cross/Dockerfile
muvaf: No, I just wanted to upgrade starting from crossplane-runtime and found this line. Should we remove it altogether if it's not used in Upbound as well?

hasheddan: @muvaf I think we need to keep the cross container around for the time-being 👍🏻

PR #190, opened 2022-06-25
Use /usr/bin/env to find bash

negz:

In some rare cases (e.g. on NixOS) bash is not at /bin/bash. Using env to find where the shell actually is is a fairly common pattern per https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability. It's also possible that /usr/bin/env is not at that path, but anecodally the internet considers it to be "more likely" to be there than /bin/bash, which is not the standard path on most BSDs and some rare Linux variants.

I have:

I've made the edit locally and run `make reviewable` against https://github.com/crossplane/crossplane.

### DIFF
--- a/makelib/common.mk
+++ b/makelib/common.mk
@@ -16,7 +16,7 @@
 .SUFFIXES:
 
 # set the shell to bash always
-SHELL := /bin/bash
+SHELL := /usr/bin/env bash
 
 # default target is build
 .PHONY: all


negz:
FWIW it's possible (but slightly annoying) to work around this by running (e.g.) `make SHELL=$(env bash) reviewable`.

hasheddan (approved):

PR #195, opened 2022-07-21
Go: Clamp down minimum required version to 1.18

epk:
Signed-off-by: Aditya Sharma <git@adi.run>

- Enforce/Require usage of Go 1.18
- Drops usage of `-i` flag for go invocations	  
  #### Why?
	  The -i flag installs the packages that are dependencies of the target.
	  The -i flag is deprecated. Compiled packages are cached automatically.

I have:

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -15,6 +15,12 @@
 # ====================================================================================
 # Options
 
+# Optional. The Go Binary to use
+GO ?= go
+
+# Optional. Minimum Go version.
+GO_REQUIRED_VERSION ?= 1.18
+
 # The go project including repo name, for example, github.com/rook/rook
 GO_PROJECT ?= $(PROJECT_REPO)
 
@@ -72,7 +78,7 @@ ifneq ($(GO_TEST_SUITE),)
 GO_TEST_FLAGS += -run '$(GO_TEST_SUITE)'
 endif
 
-GOPATH := $(shell go env GOPATH)
+GOPATH := $(shell $(GO) env GOPATH)
 
 # setup tools used during the build
 DEP_VERSION=v0.5.1
@@ -81,12 +87,10 @@ GOJUNIT := $(TOOLS_HOST_DIR)/go-junit-report
 GOCOVER_COBERTURA := $(TOOLS_HOST_DIR)/gocover-cobertura
 GOIMPORTS := $(TOOLS_HOST_DIR)/goimports
 
-GO := go
-GOHOST := GOOS=$(GOHOSTOS) GOARCH=$(GOHOSTARCH) go
+GOHOST := GOOS=$(GOHOSTOS) GOARCH=$(GOHOSTARCH) $(GO)
 GO_VERSION := $(shell $(GO) version | sed -ne 's/[^0-9]*\(\([0-9]\.\)\{0,4\}[0-9][^.]\).*/\1/p')
-GO_REQUIRED_VERSION := 1.18
 ifneq ($(GO_VERSION),$(GO_REQUIRED_VERSION))
-$(error "go version $(GO_VERSION) is too old. required version is $(GO_REQUIRED_VERSION)")
+$(error "$(GO) version $(GO_VERSION) is too old. required version is $(GO_REQUIRED_VERSION)")
 endif
 
 # we use a consistent version of gofmt even while running different go compilers.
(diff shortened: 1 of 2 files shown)


jbw976 (commented):
I cannot remember the exact scenario, but I have a vague memory recently @epk where some of our tooling, or code gen, or something wasn't compatible with go 1.18 yet.  Have you been able to use these changes to build a few different projects that consume this build submodule?

epk:
> I cannot remember the exact scenario, but I have a vague memory recently @epk where some of our tooling, or code gen, or something wasn't compatible with go 1.18 yet. Have you been able to use these changes to build a few different projects that consume this build submodule?

I have tried some repos out here: https://github.com/upbound/squad-prodeng/issues/932

muvaf:
@epk Could you test updating [`crossplane-runtime`](https://github.com/crossplane/crossplane-runtime) and [`crossplane`](https://github.com/crossplane/crossplane)? I remember some k8s libraries we use weren't compatible with go 1.18

epk:
> @epk Could you test updating [`crossplane-runtime`](https://github.com/crossplane/crossplane-runtime) and [`crossplane`](https://github.com/crossplane/crossplane)? I remember some k8s libraries we use weren't compatible with go 1.18

Opened PRs for both:
- https://github.com/crossplane/crossplane-runtime/pull/342
- https://github.com/crossplane/crossplane/pull/3200

epk:
@muvaf @jbw976 can we merge this?

makelib/golang.mk
```diff
@@ -84,10 +84,14 @@ GOIMPORTS := $(TOOLS_HOST_DIR)/goimports
 GO := go
 GOHOST := GOOS=$(GOHOSTOS) GOARCH=$(GOHOSTARCH) go
 GO_VERSION := $(shell $(GO) version | sed -ne 's/[^0-9]*\(\([0-9]\.\)\{0,4\}[0-9][^.]\).*/\1/p')
+GO_REQUIRED_VERSION := 1.18
+ifneq ($(GO_VERSION),$(GO_REQUIRED_VERSION))
```
muvaf, line 88: We used to have something like this but the friction it creates turned out to be more than the benefits we're getting for streamlining. So, I'd opt to let individual projects impose this requirement if they'd like to in their `Makefile`s.

epk: Made `GO_REQUIRED_VERSION` overridable in https://github.com/upbound/build/pull/195/commits/09af1abbfb10263202cd9a03893d63054d53b8e0

muvaf (approved):
Thanks @epk !

muvaf:
@epk I have different projects using different Go versions and I went ahead and added this override for each of them, but now I have to have a different Go binary for each version. It's possible to set up but I'm questioning the value we get here with this requirement given that Go is perfectly fine building projects with different versions and it errors out when the compatibility is broken.

Do you feel strongly about this? Is there a value I'm not seeing other than having the locally built binaries same as production? I'd like to revert it if not.

epk:
> @epk I have different projects using different Go versions and I went ahead and added this override for each of them, but now I have to have a different Go binary for each version. It's possible to set up but I'm questioning the value we get here with this requirement given that Go is perfectly fine building projects with different versions and it errors out when the compatibility is broken.
> 
> 
> 
> Do you feel strongly about this? Is there a value I'm not seeing other than having the locally built binaries same as production? I'd like to revert it if not.

You can use https://github.com/travis-ci/gimme alongside dotenv to automatically load the go versions are required. 

I'll comment back tomorrow but several upstream K8s dependencies also now strictly require 1.18 because of the usage of `any`.

muvaf:
> I'll comment back tomorrow but several upstream K8s dependencies also now strictly require 1.18 because of the usage of any.

FWIW, when Go 1.17 sees `any` it returns error just like the one we have here anyway, so it already forces you to upgrade. The thing is the whole transition cost will be needed to be paid for the next versions as well since we're forcing a specific version even though it may not really be required by the toolchain and `gimme` helps but it's yet another tool developers need to know about. All of it can be worked out and let's keep it if you feel strongly about it, but I just want us to consider what we're getting in return. Is it reproducibility or implicit push for owners to upgrade to the latest version or something else?

turkenh:
This check is causing trouble and confusion since it is being invoked even if you don't call a target building go code.
See: https://github.com/upbound/uptest/pull/39#discussion_r1010456571

PR #196, opened 2022-07-27
Go: miscellaneous cleanup

epk:
Signed-off-by: Aditya Sharma <git@adi.run>

### Description of your changes

Cleans up some of the stuff in the Go build module that is no longer relevant. Inline comments follow

I have:

- [X] Read and followed Upbound's [contribution process](https://git.io/fj2m9).
- [ ] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR, as appropriate.

### How has this code been tested

Tested all `go` targets with `managed-control-plane` and `shimmer`

PR #201, opened 2022-08-05
Better error message for `go.modules.tidy.check` failures

epk:

Changes the error to be more verbose with instructions

I have:

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -197,7 +197,7 @@ go.fmt.simplify: $(GOFMT)
 	@$(GOFMT) -l -s -w $(GO_SUBDIRS) $(GO_INTEGRATION_TESTS_SUBDIRS) || $(FAIL)
 	@$(OK) gofmt simplify
 
-go.validate: go.vet go.fmt
+go.validate: go.modules.check go.vet go.fmt
 
 go.vendor.lite: go.modules.verify
 go.vendor.check: go.modules.check
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -199,7 +199,8 @@ go.fmt.simplify: $(GOFMT)
 
 go.validate: go.vet go.fmt
 
-go.vendor.lite go.vendor.check: go.modules.check
+go.vendor.lite: go.modules.verify
+go.vendor.check: go.modules.check
 go.vendor.update: go.modules.update
 go.vendor: go.modules.download
 
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -223,7 +223,7 @@ go.modules.tidy:
 go.modules.tidy.check:
 	@$(INFO) verify go modules dependencies are tidy
 	@$(GO) mod tidy
-	@changed=$$(git diff --exit-code --name-only go.mod go.sum 2>&1) && [ -z "$${changed}" ] || (echo "go.mod is not tidy" 1>&2; $(FAIL))
+	@changed=$$(git diff --exit-code --name-only go.mod go.sum 2>&1) && [ -z "$${changed}" ] || (echo "go.mod is not tidy. Please run 'make go.modules.tidy' and stage the changes" 1>&2; $(FAIL))
 	@$(OK) go modules are tidy
 
 go.modules.update:


makelib/golang.mk
```diff
@@ -223,7 +223,7 @@ go.modules.tidy:
 go.modules.tidy.check:
 	@$(INFO) verify go modules dependencies are tidy
 	@$(GO) mod tidy
-	@changed=$$(git diff --exit-code --name-only go.mod go.sum 2>&1) && [ -z "$${changed}" ] || (echo "go.mod is not tidy" 1>&2; $(FAIL))
+	@changed=$$(git diff --exit-code --name-only go.mod go.sum 2>&1) && [ -z "$${changed}" ] || (echo "go.mod is not tidy. Please run 'make go.modules.tidy' and stage the changes" 1>&2; $(FAIL))
```
muvaf, line 226: Is there a case where this check is useful? We run `go mod tidy` before doing the check and if it's successful, then we are in the tidy state; if it fails, then it'd return its own error anyway telling us it's not tidy at the moment.

epk: `go mod tidy` does not return an error at the moment.

epk: Oh looks like there's a flag `-v`:
```
go help mod tidy
usage: go mod tidy [-e] [-v] [-go=version] [-compat=version]

Tidy makes sure go.mod matches the source code in the module.
It adds any missing modules necessary to build the current module's
packages and dependencies, and it removes unused modules that
don't provide any relevant packages. It also adds any missing entries
to go.sum and removes any unnecessary ones.

The -v flag causes tidy to print information about removed modules
to standard error.

The -e flag causes tidy to attempt to proceed despite errors
encountered while loading packages.

The -go flag causes tidy to update the 'go' directive in the go.mod
file to the given version, which may change which module dependencies
are retained as explicit requirements in the go.mod file.
(Go versions 1.17 and higher retain more requirements in order to
support lazy module loading.)

The -compat flag preserves any additional checksums needed for the
'go' command from the indicated major Go release to successfully load
the module graph, and causes tidy to error out if that version of the
'go' command would load any imported package from a different module
version. By default, tidy acts as if the -compat flag were set to the
version prior to the one indicated by the 'go' directive in the go.mod
file.

See https://golang.org/ref/mod#go-mod-tidy for more about 'go mod tidy'.
```

epk: `-v` however does not error out for additions and or dependencies changes (ex: an indirect dependency becoming direct)

epk: Relevant issue: https://github.com/golang/go/issues/27005

muvaf: > go mod tidy does not return an error at the moment.

It'd actually return error when it couldn't download the module or somehow there is a conflict in `go.sum`, right?

What I'm sort of questioning is whether the module being tidy or not is important once `go mod tidy` passes successfully. For diff check, `make check-diff` already checks whether you need something auto-generated to be committed or not. So, I'd propose to remove this line altogether.

epk: `make check-diff` would need to explicitly call `go.modules.tidy` as that only runs `make generate`

I'll add this there, and change `vendor.lite` to only call `go.modules.verify`

epk: Locally tidiness is not important but we shouldn't be opening PRs with an untidy go.mod

epk: Restored the older behaviour in [`cf87c7e` (#201)](https://github.com/upbound/build/pull/201/commits/cf87c7ec9b0cea2fdac6f709d03690253c54e823):

- `go.init` calls `go.vendor.lite`
- `go.vendor.lite` calls `go.modules.download go.modules.verify`

This now matches the behaviour before https://github.com/upbound/build/pull/196, where `go.vendor.lite` used to call `go mod verify`: 
```
go.vendor.lite go.vendor.check:
 	@$(INFO) verify dependencies have expected content
 	@$(GOHOST) mod verify || $(FAIL)
 	@$(OK) go modules dependencies verified
```

muvaf: @epk thank you!

muvaf (approved):
LGTM!

PR #205, opened 2022-09-14
fix:(helm): Set HELM_*_HOME env vars to avoid side effects with go

MisterMX:

This removes exports of the `XDG_*` environment variables and exports `HELM_*_HOME` instead which has the same effect in Helm but avoids the side effects with Go that causes `GOCACHE` to be set to `.work/go-build`.

I have:

Manually


negz (approved):

turkenh (approved):
Thanks @MisterMX 🙌

PR #207, opened 2022-09-29
Pass controller flag only for provider packages

turkenh:

To make recently introduced xpkg build functionality work with configuration packages as well, we will be passing `--controller` argument to xpkg build command only for provider packages.

I have:

Together with [platform-ref-gcp](https://github.com/upbound/platform-ref-gcp) with `make build` using the following Makefile:

```
# Project Setup
PROJECT_NAME := platform-ref-gcp
PROJECT_REPO := github.com/upbound/$(PROJECT_NAME)

PLATFORMS ?= linux_amd64 linux_arm64
include build/makelib/common.mk

# ====================================================================================
# Setup Kubernetes tools

UP_VERSION = v0.13.0
UP_CHANNEL = stable

-include build/makelib/k8s_tools.mk
# ====================================================================================
# Setup XPKG

XPKG_REG_ORGS ?= xpkg.upbound.io/upbound
# NOTE(hasheddan): skip promoting on xpkg.upbound.io as channel tags are
# inferred.
XPKG_REG_ORGS_NO_PROMOTE ?= xpkg.upbound.io/upbound
XPKGS = $(PROJECT_NAME)
-include build/makelib/xpkg.mk

# ====================================================================================
# Targets

# NOTE(hasheddan): we must ensure up is installed in tool cache prior to build
# as including the k8s_tools machinery prior to the xpkg machinery sets UP to
# point to tool cache.
build.init: $(UP)
```

### DIFF
--- a/makelib/xpkg.mk
+++ b/makelib/xpkg.mk
@@ -64,8 +64,9 @@ define xpkg.build.targets
 xpkg.build.$(1):
 	@$(INFO) Building package $(1)-$(VERSION).xpkg for $(PLATFORM)
 	@mkdir -p $(OUTPUT_DIR)/xpkg/$(PLATFORM)
-	@$(UP) xpkg build \
-		--controller $(BUILD_REGISTRY)/$(1)-$(ARCH) \
+	@controller_arg=$$$$(grep -E '^kind:\s+Provider\s*$$$$' $(XPKG_DIR)/crossplane.yaml > /dev/null && echo "--controller $(BUILD_REGISTRY)/$(1)-$(ARCH)"); \
+	$(UP) xpkg build \
+		$$$${controller_arg} \
 		--package-root $(XPKG_DIR) \
 		--examples-root $(XPKG_EXAMPLES_DIR) \
 		--ignore $(XPKG_IGNORE) \


makelib/xpkg.mk
```diff
@@ -64,8 +64,9 @@ define xpkg.build.targets
 xpkg.build.$(1):
 	@$(INFO) Building package $(1)-$(VERSION).xpkg for $(PLATFORM)
 	@mkdir -p $(OUTPUT_DIR)/xpkg/$(PLATFORM)
-	@$(UP) xpkg build \
-		--controller $(BUILD_REGISTRY)/$(1)-$(ARCH) \
+	@controller_arg=$$$$(grep -E '^kind:\s+Provider\s*$$$$' $(XPKG_DIR)/crossplane.yaml > /dev/null && echo "--controller $(BUILD_REGISTRY)/$(1)-$(ARCH)"); \
```
hasheddan, line 67: I’m fine with doing this for now, but ultimately the fact that the kind is a Provider does not necessarily indicate that we want to specify a base controller image. The other option I was exploring was a separate list of values denoting which packages required a controller reference, which would be more work for the consumer of this package but also less opaque. Longer term I anticipate using a dedicated GitHub action for all of this machinery. Given the alternative solution, would you still prefer to go with solution provided here?

turkenh: yeah, I agree that this should not be the ultimate solution and think this as a stopgap solution. 

> the fact that the kind is a Provider does not necessarily indicate that we want to specify a base controller image

which is already the case, so this is PR is not changing something in that regard.

> Given the alternative solution, would you still prefer to go with solution provided here?

That was something I considered as well, but didn't want to build that since I didn't feel comfortable with keeping two lists. So wanted make a minor change to unblock myself without making things worse. 
So, yes. I would like to proceed if you agree with the above.

hasheddan: That seems reasonable to me 👍🏻

hasheddan (approved):

makelib/xpkg.mk
turkenh: > Longer term I anticipate using a dedicated GitHub action for all of this machinery.

As a side note, I see some value with having in the build submodule compared to having it as a GitHub action like trying/testing things locally.

PR #208, opened 2022-09-29
Add targets to deploy a locally build configuration package

turkenh:

Adds targets to deploy a locally build configuration package.

It also adds targets to start and stop a control plane (a.k.a. K8s + XP)

I have:

With this PR: https://github.com/upbound/platform-ref-gcp/pull/25
See `uptest` target.

### DIFF
--- a/makelib/local.xpkg.mk
+++ b/makelib/local.xpkg.mk
@@ -31,7 +31,7 @@ local.xpkg.sync: local.xpkg.init $(UP)
 		$(KUBECTL) -n $(CROSSPLANE_NAMESPACE) cp $(XPKG_OUTPUT_DIR)/cache -c dev $$XPPOD:/tmp
 	@$(OK) copying local xpkg cache to Crossplane pod
 
-local.xpkg.deploy.%: local.xpkg.sync
-	@$(INFO) deploying xpkg $* $(VERSION)
+local.xpkg.deploy-cfg.%: local.xpkg.sync
+	@$(INFO) deploying configuration package $* $(VERSION)
 	@echo '{"apiVersion":"pkg.crossplane.io/v1","kind":"Configuration","metadata":{"name":"$*"},"spec":{"package":"$*-$(VERSION).gz","packagePullPolicy":"Never"}}' | $(KUBECTL) apply -f -
-	@$(OK) deploying xpkg $* $(VERSION)
+	@$(OK) deploying configuration package $* $(VERSION)
--- a/makelib/local.xpkg.mk
+++ b/makelib/local.xpkg.mk
@@ -1,0 +1,37 @@
+# Copyright 2022 The Upbound Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+CROSSPLANE_NAMESPACE ?= crossplane-system
+
+local.xpkg.init: $(KUBECTL)
+	@$(INFO) patching Crossplane with dev sidecar
+	@if ! $(KUBECTL) -n $(CROSSPLANE_NAMESPACE) get deployment crossplane -o jsonpath="{.spec.template.spec.containers[*].name}" | grep "dev" > /dev/null; then \
+		$(KUBECTL) -n $(CROSSPLANE_NAMESPACE) patch deployment/crossplane --type='json' -p='[{"op":"add","path":"/spec/template/spec/containers/1","value":{"image":"nginx","imagePullPolicy":"Always","name":"dev","resources":{},"terminationMessagePath":"/dev/termination-log","terminationMessagePolicy":"File","volumeMounts":[{"mountPath":"/tmp/cache","name":"package-cache"}]}},{"op":"add","path":"/spec/template/metadata/labels/patched","value":"true"}]'; \
+		$(KUBECTL) -n $(CROSSPLANE_NAMESPACE) wait deploy crossplane --for condition=Available --timeout=60s; \
+		$(KUBECTL) -n $(CROSSPLANE_NAMESPACE) wait pods -l app=crossplane,patched=true --for condition=Ready --timeout=60s; \
+	fi
+	@$(OK) patching Crossplane with dev sidecar
+
+local.xpkg.sync: local.xpkg.init $(UP)
+	@$(INFO) copying local xpkg cache to Crossplane pod
+	@mkdir -p $(XPKG_OUTPUT_DIR)/cache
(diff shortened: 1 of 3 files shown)


makelib/local.xpkg.mk
```diff
@@ -0,0 +1,37 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+CROSSPLANE_NAMESPACE ?= crossplane-system
+
+local.xpkg.init: $(KUBECTL)
+	@$(INFO) patching Crossplane with dev sidecar
+	@if ! $(KUBECTL) -n $(CROSSPLANE_NAMESPACE) get deployment crossplane -o jsonpath="{.spec.template.spec.containers[*].name}" | grep "dev" > /dev/null; then \
+		$(KUBECTL) -n $(CROSSPLANE_NAMESPACE) patch deployment/crossplane --type='json' -p='[{"op":"add","path":"/spec/template/spec/containers/1","value":{"image":"nginx","imagePullPolicy":"Always","name":"dev","resources":{},"terminationMessagePath":"/dev/termination-log","terminationMessagePolicy":"File","volumeMounts":[{"mountPath":"/tmp/cache","name":"package-cache"}]}},{"op":"add","path":"/spec/template/metadata/labels/patched","value":"true"}]'; \
```
muvaf, line 20: Is there a specific reason to use `nginx` instead of alpine or busybox?

makelib/local.xpkg.mk
```diff
@@ -0,0 +1,37 @@
...
+	@$(INFO) copying local xpkg cache to Crossplane pod
+	@mkdir -p $(XPKG_OUTPUT_DIR)/cache
+	@for pkg in $(XPKG_OUTPUT_DIR)/linux_*/*; do $(UP) xpkg xp-extract --from-xpkg $$pkg -o $(XPKG_OUTPUT_DIR)/cache/$$(basename $$pkg .xpkg).gz; done
+	@XPPOD=$$($(KUBECTL) -n $(CROSSPLANE_NAMESPACE) get pod -l app=crossplane,patched=true -o jsonpath="{.items[0].metadata.name}"); \
+		$(KUBECTL) -n $(CROSSPLANE_NAMESPACE) cp $(XPKG_OUTPUT_DIR)/cache -c dev $$XPPOD:/tmp
+	@$(OK) copying local xpkg cache to Crossplane pod
+
+local.xpkg.deploy-cfg.%: local.xpkg.sync
+	@$(INFO) deploying configuration package $* $(VERSION)
+	@echo '{"apiVersion":"pkg.crossplane.io/v1","kind":"Configuration","metadata":{"name":"$*"},"spec":{"package":"$*-$(VERSION).gz","packagePullPolicy":"Never"}}' | $(KUBECTL) apply -f -
```
muvaf, line 36: I think giving the package name without a tag should be sufficient. [Here](https://github.com/upbound/provider-aws/blob/main/cluster/install_provider.sh#L141) is how that's used in providers.

makelib/local.xpkg.mk
```diff
@@ -0,0 +1,37 @@
...
+local.xpkg.init: $(KUBECTL)
+	@$(INFO) patching Crossplane with dev sidecar
+	@if ! $(KUBECTL) -n $(CROSSPLANE_NAMESPACE) get deployment crossplane -o jsonpath="{.spec.template.spec.containers[*].name}" | grep "dev" > /dev/null; then \
+		$(KUBECTL) -n $(CROSSPLANE_NAMESPACE) patch deployment/crossplane --type='json' -p='[{"op":"add","path":"/spec/template/spec/containers/1","value":{"image":"nginx","imagePullPolicy":"Always","name":"dev","resources":{},"terminationMessagePath":"/dev/termination-log","terminationMessagePolicy":"File","volumeMounts":[{"mountPath":"/tmp/cache","name":"package-cache"}]}},{"op":"add","path":"/spec/template/metadata/labels/patched","value":"true"}]'; \
+		$(KUBECTL) -n $(CROSSPLANE_NAMESPACE) wait deploy crossplane --for condition=Available --timeout=60s; \
+		$(KUBECTL) -n $(CROSSPLANE_NAMESPACE) wait pods -l app=crossplane,patched=true --for condition=Ready --timeout=60s; \
+	fi
+	@$(OK) patching Crossplane with dev sidecar
+
+local.xpkg.sync: local.xpkg.init $(UP)
```
muvaf, line 26:
```suggestion
local.xpkg.cache: local.xpkg.init $(UP)
```

nitpick

makelib/local.xpkg.mk
```diff
@@ -0,0 +1,37 @@
...
+	@$(INFO) copying local xpkg cache to Crossplane pod
+	@mkdir -p $(XPKG_OUTPUT_DIR)/cache
+	@for pkg in $(XPKG_OUTPUT_DIR)/linux_*/*; do $(UP) xpkg xp-extract --from-xpkg $$pkg -o $(XPKG_OUTPUT_DIR)/cache/$$(basename $$pkg .xpkg).gz; done
+	@XPPOD=$$($(KUBECTL) -n $(CROSSPLANE_NAMESPACE) get pod -l app=crossplane,patched=true -o jsonpath="{.items[0].metadata.name}"); \
+		$(KUBECTL) -n $(CROSSPLANE_NAMESPACE) cp $(XPKG_OUTPUT_DIR)/cache -c dev $$XPPOD:/tmp
+	@$(OK) copying local xpkg cache to Crossplane pod
+
+local.xpkg.deploy-cfg.%: local.xpkg.sync
```
muvaf, line 34:
```suggestion
local.xpkg.deploy.configuration.%: local.xpkg.sync
```

nitpick

muvaf (approved):
Thanks @turkenh !

makelib/local.xpkg.mk
```diff
@@ -0,0 +1,37 @@
...
+	@$(INFO) deploying configuration package $* $(VERSION)
```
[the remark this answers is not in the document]
muvaf: FWIW, since we don't do any building here, `xpkg` parts are same for both types.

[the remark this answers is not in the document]
turkenh: @muvaf the issue with providers is making its controller image available to the local cluster. 
You're right, it would deploy provider pkg same as this one but things get complicated when I wanted to make provider image available using the same flow. 

I can give it another try later though to find a way otherwise share the exact problem.

makelib/local.xpkg.mk
turkenh: initially used since it keeps the debug container running without additonal commands but using alpine or others with some sleep command makes sense as well. I'll check that.

makelib/local.xpkg.mk
turkenh: I think it is related how it was named during extraction [here](https://github.com/upbound/provider-aws/blob/ba4e3f989891df3f9f7786b25992c0f884fede74/cluster/install_provider.sh#L53).

I wanted to include the version since it helps to iterate during development, e.g. make a change, commit, deploy and wait your changes to be reflected on a running cp.

makelib/local.xpkg.mk
turkenh: It synchronizes local cache directory to the Crossplane pods. 
I could use `local.xpkg.sync-cache` but prefer to keep it as is to brevity.

PR #224, opened 2022-12-13
fix(docs): Run npm install on docs.init

MisterMX:

Fix error in pipeline due to missing node packages, such as

```
Error: Error building site: POSTCSS: failed to transform "scss/docs.css" (text/css). Check your PostCSS installation; install with "npm install postcss-cli". See https://gohugo.io/hugo-pipes/postcss/: this feature is not available in your current Hugo version, see https://goo.gl/YMrWcn for more information
```

I have:

Manually

### DIFF
--- a/makelib/docs.mk
+++ b/makelib/docs.mk
@@ -42,6 +42,7 @@ docs.init:
 	rm -rf $(DOCS_WORK_DIR)
 	mkdir -p $(DOCS_WORK_DIR)
 	git clone --depth=1 -b master $(DOCS_GIT_REPO) $(DOCS_WORK_DIR)
+	(cd $(DOCS_WORK_DIR) && $(MAKE) npm.install)
 
 docs.generate: docs.init
 	rm -rf $(DOCS_VERSION_DIR)


plumbis:
Fixes [docs issue #214](https://github.com/crossplane/docs/issues/214)

makelib/docs.mk
```diff
@@ -42,6 +42,7 @@ docs.init:
 	rm -rf $(DOCS_WORK_DIR)
 	mkdir -p $(DOCS_WORK_DIR)
 	git clone --depth=1 -b master $(DOCS_GIT_REPO) $(DOCS_WORK_DIR)
+	(cd $(DOCS_WORK_DIR) && npm install)
```
hasheddan, line 45: This may not be required in all cases -- would suggest adding as a dependency to targets in https://github.com/crossplane/docs/blob/9fd142c5c855630dd8a55a4a19194428243a42b6/Makefile#L76

MisterMX: I've added an `npm.install` make target to the docs in https://github.com/crossplane/docs/pull/233. And called `make npm.install` here instead.

It is needed here because `docs.init` prunes and recreates the docs repo so all installed are packages are lost, too.

haarchri:
Closing as not needed since Make was removed from the docs repo.

PR #225, opened 2022-12-16
Remove docs publishing and version-specific changes

plumbis:
This removes `docs.publish` and all version related actions for docs. 

This is related to crossplane/crossplane issue [#3461](https://github.com/crossplane/crossplane/issues/3461), moving the docs out of the crossplane/crossplane repo. 

Other docs make options are left in case a crossplane developer need to build docs locally for some reason. 

Signed-off-by: Pete Lumbis <pete@upbound.io>

### DIFF
--- a/makelib/docs.mk
+++ b/makelib/docs.mk
@@ -15,25 +15,11 @@
 # ====================================================================================
 # Options
 
-ifndef SOURCE_DOCS_DIR
-$(error SOURCE_DOCS_DIR must be defined)
-endif
-
-ifndef DEST_DOCS_DIR
-$(error DEST_DOCS_DIR must be defined)
-endif
-
 ifndef DOCS_GIT_REPO
 $(error DOCS_GIT_REPO must be defined)
 endif
 
-# Optional. If false the publish step will remove this version from the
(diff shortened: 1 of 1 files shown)


hasheddan (changes requested):
@plumbis if this is the direction we end up going I think we can do it without ripping out this generic functionality (i.e. make it so crossplane doesn't publish docs rather than make it so that the build submodule can't publish docs).

plumbis:
This method was to split the difference and still enable local docs builds/previews from crossplane (see Crossplane [PR #3538](https://github.com/crossplane/crossplane/pull/3538)) 

If we want to remove all docs related things from Crossplane then I'm fine with closing this.

hasheddan:
> This method was to split the difference and still enable local docs builds/previews from crossplane

@plumbis my point is that this PR is removing common functionality. If Crossplane wants to do local builds and previews, but _not_ publish, we should enable that in this module, rather than removing the ability to publish altogether.

plumbis:
Thanks Dan. I'll close this and make changes to the Makefiles in the other repos.

PR #226, opened 2023-03-01
Bump version of golangci-lint to v1.51.0

siredmar:
Fixes https://github.com/golangci/golangci-lint/issues/3565

### DIFF
--- a/makelib/golang.mk
+++ b/makelib/golang.mk
@@ -98,7 +98,7 @@ endif
 
 # We use a consistent version of golangci-lint to ensure everyone gets the same
 # linters.
-GOLANGCILINT_VERSION ?= 1.50.1
+GOLANGCILINT_VERSION ?= 1.51.0
 GOLANGCILINT := $(TOOLS_HOST_DIR)/golangci-lint-v$(GOLANGCILINT_VERSION)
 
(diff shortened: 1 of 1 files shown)


siredmar:
@turkenh would you mind reviewing this PR?

makelib/golang.mk
```diff
@@ -98,7 +98,7 @@ endif
 
 # We use a consistent version of golangci-lint to ensure everyone gets the same
 # linters.
-GOLANGCILINT_VERSION ?= 1.50.1
+GOLANGCILINT_VERSION ?= 1.51.0
```
hasheddan, line 101: Thanks @siredmar! If you are encountering this in a repo you can always override in your parent `Makefile` by explicitly setting `GOLANGCILINT_VERSION` as well :) Also, any reason not to go ahead and bump to latest (`1.52.2`)?

siredmar: I guess the reason for this was the lined version in the issue mentioned above. I didn't test go1.20 with 1.52.2. If you like, you can test it as i don't have the time to test this right now.

hasheddan (approved):
This will get us past the linked bug — we can bump to a newer version as needed. Thanks @siredmar!

PR #249, opened 2024-01-16
Add yq As a Tool

ulucinar:

We are planning to use `kustomize` to process the family provider CRDs. `kustomize` outputs a multi-document YAML that contains all the processed CRDs and the `up xpkg batch` command relies on the controller-tools naming conventions for the CRD files to decide which CRD goes into which family provider package. Thus it cannot currently partition CRDs if they are in a single file. I checked the possibility of a content based filter (instead of relying on the generated file name) but the  interface is [`io.ReadCloser`](https://pkg.go.dev/io#ReadCloser), which processes a byte stream and is not aware of the YAML documents. Implementing a content filter will introduce unnecessary complexity due to the mismatching abstraction layers. Also considered a filtering [`parser.Parser`](https://pkg.go.dev/github.com/crossplane/crossplane-runtime/pkg/parser#Parser) implementation but that interface returns a concrete `parser.Package` (struct), which does not allow writes to its content list. So, that approach would require us to change the `parser.Parser` interface.

Instead of introducing `yq`, we first implemented a bash script processing the multi-doc YAML file line-by-line and splitting the over 900 CRDs of the official AWS provider family and it takes ~70 s on an M2 Mac to do so. Doing experiments with `yq`, it completes in ~10 s. So, we would like to use `yq` for splitting the multi-doc YAML output from `kustomize`.

I have:

Tested both on the darwin ARM64 & linux AMD64 platforms.

### DIFF
--- a/makelib/k8s_tools.mk
+++ b/makelib/k8s_tools.mk
@@ -71,6 +71,11 @@ KUTTL := $(TOOLS_HOST_DIR)/kuttl-$(KUTTL_VERSION)
 # the version of uptest to use
 UPTEST_VERSION ?= v0.1.0
 UPTEST := $(TOOLS_HOST_DIR)/uptest-$(UPTEST_VERSION)
+
+# the version of yq to use
+YQ_VERSION ?= v4.40.5
+YQ := $(TOOLS_HOST_DIR)/yq-$(YQ_VERSION)
+
 # ====================================================================================
 # Common Targets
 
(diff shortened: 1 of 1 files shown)


sergenyalcin (approved):
Thanks @ulucinar LGTM!

turkenh (approved):
Thanks @ulucinar 🙌

PR #255, opened 2024-04-19
Update download URL of uptest

sergenyalcin:

Change the download URL of uptest.

I have:

Tested locally

### DIFF
--- a/makelib/k8s_tools.mk
+++ b/makelib/k8s_tools.mk
@@ -70,6 +70,7 @@ KUTTL := $(TOOLS_HOST_DIR)/kuttl-$(KUTTL_VERSION)
 
 # the version of uptest to use
 UPTEST_VERSION ?= v0.1.0
+UPTEST_CHANNEL ?= stable
 UPTEST := $(TOOLS_HOST_DIR)/uptest-$(UPTEST_VERSION)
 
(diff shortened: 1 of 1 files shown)


ulucinar (approved):
Thanks @sergenyalcin, lgtm. Btw, after this PR is merged, build pipelines depending on an uptest version which is not present at the new location will break. Either we make this clear so that users of uptest update the version at least to `v0.12.0` when they bump their repo's build submodules version, or if we want to be super kind to them, we could also push the previous releases of uptest to the new location so that upgrades are smoother. I currently assume this repo is mostly used in the Upbound organization but it's also a public repo (which might be a motivation for some extra care).

makelib/k8s_tools.mk
```diff
@@ -70,6 +70,7 @@ KUTTL := $(TOOLS_HOST_DIR)/kuttl-$(KUTTL_VERSION)
 
 # the version of uptest to use
 UPTEST_VERSION ?= v0.1.0
```
ulucinar, line 72: Let's also update the default version to the min version available at the new location, e.g., `v0.12.0.

sergenyalcin: Done!

